PR #6, opened 2023-07-11
Add breaking change regarding the docker image generation of APIM 

phiz71:
This PR adds sections in the BC part of APIM changelog, just to make this doc a little easier to read.

this PR also adds a new BC regarding the way we generate our docker images.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .DS_Store
 .env
 .env.*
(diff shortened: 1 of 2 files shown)


phiz71:
@Lyd1aCla1r3 @bigmike36c,
I wanted to add both of you as reviewers, but I can only set one.

PR #62, opened 2023-11-20
Fix docker compose for Consul

jmcx:
The Consul docker compose snippet shown here refers to a network called "backend", but in the main docker compose file recommended by the doc, the two networks defined are "frontend" and "storage". Assuming here that it should be "storage" for Consul, although I am testing now.

image


Lyd1aCla1r3:
Hi @jmcx, are we good to go with "storage" following your test?

Lyd1aCla1r3 (approved):

PR #78, opened 2023-12-21
fix(am-1175): added missing AM API ref link

ashraf706:
Added AM API missing links for 4.2,4.1, 4.0


leleueri (commented):
**issue**: Don't think we have to reference the legacy docs website for 4.x specification. Starting from 4.1, we are generating openapi.yaml and I don't know if the "online reference" will interpret the content properly. IMO we should provide the specification for download only. The spec should be committed in the [AM repository](https://github.com/gravitee-io/gravitee-access-management/tree/master/docs) with the specification version consistent with the branch name.

Lyd1aCla1r3 (approved):

PR #169, opened 2024-05-17
doc: consistent urls in the example

SelviA:
modifying the urls in the examples to be consistent

Entrypoint in the first example is `/qs-traditional-api` and should be the same in the *Example continued* section.


bigmike36c (approved):
Looks good! Thanks for the PR!

Lyd1aCla1r3 (approved):

PR #289, opened 2024-09-24
docs: add a note about api-key usage in expression language for APIs using v4 emulation engine

Okhelifi:
## Description

In this PR we add a small note about the usage of the `api-key` attribute for APIs using v4 emulation engine.
Basically to resume it, something was working before v4 and does not work now with V4.

### DIFF
--- a/docs/apim/4.0/guides/gravitee-expression-language.md
+++ b/docs/apim/4.0/guides/gravitee-expression-language.md
@@ -193,6 +193,17 @@ Request context attributes and examples are listed below.
 {% endtab %}
 {% endtabs %}
 
+#### Important Note for v4 Emulation Engine Users
+When you use the v4 emulation engine, you retrieve the API Key from the request parameters or headers. You cannot reference the API directly with expressions like `{#context.attributes['api-key'] == 'api-key-value'}` because the api-key attribute is not accessible in the context as a logging filter expression.
+
+To retrieve API key from the request parameters or headers, use the following expression:
+
+```
+{(#request.params['api-key'] != null && #request.params['api-key'][0] == 'my-api-key') || #request.headers['X-Gravitee-Api-Key'][0] == 'my-api-key'}
+```
+
+This approach checks for the API key either in the query parameters (api-key) or in the headers (X-Gravitee-Api-Key), which ensures compatibility with v4 emulation.
+
 ### SSL object properties <a href="#ssl_object" id="ssl_object"></a>
 
 The object properties you can access in the `ssl` session object from the `{#request.ssl}` root-level object property are listed below.
--- a/docs/apim/4.1/guides/gravitee-expression-language.md
+++ b/docs/apim/4.1/guides/gravitee-expression-language.md
@@ -210,6 +210,15 @@ Request context attributes and examples are listed below.
 {% endtab %}
 {% endtabs %}
 
+#### Important Note for v4 Emulation Engine Users
+When you use the v4 emulation engine, you retrieve the API Key from the request parameters or headers. You cannot reference the API directly with expressions like `{#context.attributes['api-key'] == 'api-key-value'}` because the api-key attribute is not accessible in the context as a logging filter expression.
+
+```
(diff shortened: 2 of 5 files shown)


docs/apim/4.0/guides/gravitee-expression-language.md
```diff
@@ -193,6 +193,17 @@ Request context attributes and examples are listed below.
 {% endtab %}
 {% endtabs %}
 
+#### Important Note for v4 Emulation Engine Users
+When using the v4 emulation engine, it is important to note that the api-key attribute is no longer accessible in the context if you wish to use it as a logging filter condition. Instead of referencing the api-key attribute directly with expressions like `{#context.attributes['api-key'] == 'api-key-value'}`, users can retrieve the API key from request parameters or headers.
```
garethbrinn: Some changes:
When you use the v4 emulation engine, you retrieve the API key from the request parameters or headers instead of referencing the API directly with expressions like `{#context.attributes['api-key'] == 'api-key-value'}`.

docs/apim/4.0/guides/gravitee-expression-language.md
```diff
@@ -193,6 +193,17 @@ Request context attributes and examples are listed below.
...
+
+To achieve this, consider using the following expression:
```
garethbrinn: Instead of "To achieve this, consider using the following expression", I would say: "To retrieve API key from the request parameters or headers, use the following expression:

docs/apim/4.0/guides/gravitee-expression-language.md
```diff
@@ -193,6 +193,17 @@ Request context attributes and examples are listed below.
...
+
+```
+{(#request.params['api-key'] != null && #request.params['api-key'][0] == 'my-api-key') || #request.headers['X-Gravitee-Api-Key'][0] == 'my-api-key'}
+```
+
+This approach checks for the API key either in the query parameters (api-key) or in the headers (X-Gravitee-Api-Key), ensuring compatibility with v4 emulation.
```
garethbrinn: I would change "ensuring compatibility with v4 emulation." to "which ensures compatibility with v4 emulation."

docs/apim/4.0/guides/gravitee-expression-language.md
Okhelifi: Can we keep my explanation please ? It is really clearer I think from a user point of view

docs/apim/4.0/guides/gravitee-expression-language.md
Okhelifi: Done

docs/apim/4.0/guides/gravitee-expression-language.md
Okhelifi: Done

docs/apim/4.0/guides/gravitee-expression-language.md
garethbrinn: It has to be rephrased to tell the user what the user can do rather than what the user can't do. There is no point in telling the user what they can't do since the goal is to tell the user how to do something. 

What about this note: 

When you use the v4 emulation engine, you retrieve the API Key from the request parameters or headers. You cannot reference the API directly with expressions like {#context.attributes['api-key'] == 'api-key-value'} because the api-key attribute is not accessible in the context as a logging filter expression.

Okhelifi: Tbh I'm still not convinced but I'm not a native english speaker so I believe in you 😅
Done ✅

garethbrinn: If I'm honest, that makes your opinion more valuable because our documentation needs to be accessible to non-native English speakers. 

If you want, we can hop on a quick call sometime to discuss it?

Okhelifi:
Hello @garethbrinn do you think that we can approve and merge this PR. I think that all the requested changes has been done

garethbrinn (approved):

PR #302, opened 2024-10-03
feat: add a note about new USER_TOKEN permission in upgrade guides

Okhelifi:
## Description

Following this issue https://gravitee.atlassian.net/browse/APIM-6960 I'm adding a note in the upgrade guides for the incoming versions of APIM.
This PR explain to the user the creation of a new ROLE named USER_TOKEN. 

☢️ We need to merge this PR only when the issue is **done** and **the associated releases are done** ☢️

### DIFF
--- a/docs/apim/4.1/getting-started/install-guides/installation-guide-migration.md
+++ b/docs/apim/4.1/getting-started/install-guides/installation-guide-migration.md
@@ -160,3 +160,11 @@ eventsCollection.find({"type": "PUBLISH_API"}).forEach((event) => {
        eventsCollection.replaceOne({ _id: event._id }, event);
 });
 ```
+
+## Introduction of USER_TOKEN Permission for Token Management
+
+Starting with version 4.1.26, there is a new permission called USER_TOKEN. This permission controls which users have the ability to read, create, update, and delete user tokens.
+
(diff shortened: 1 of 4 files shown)


docs/apim/4.1/getting-started/install-guides/installation-guide-migration.md
```diff
@@ -160,3 +160,11 @@ eventsCollection.find({"type": "PUBLISH_API"}).forEach((event) => {
        eventsCollection.replaceOne({ _id: event._id }, event);
 });
 ```
+
+## Introduction of USER_TOKEN Permission for Token Management
+
+Starting with version 4.1.26, a new permission called 'USER_TOKEN' has been introduced. This permission is specifically designed to control which users have the ability to read, create, update, and delete user tokens.
```
garethbrinn: I would change the opening sentence to: "Starting with version 4.1.26, There is a new permission called `USER_TOKEN`. 

For the 2nd sentence, I would just say "This permission controls which...."

Okhelifi: Should be ok now

phiz71 (approved):

PR #681, opened 2025-06-20
Correct "Resource filtering" policy example for v2 API definition

ryancabanas:
- Correctly set the `"policy"` property to `"resource-filtering"` in the example for the "v2 API definition" section of the "Resource Filtering" policy documentation.


ryancabanas:
Bad change?  Rebase?

PR #781, opened 2025-09-22
docs: add note about IPV4 only for APIM frontend deployments

a-cordier:


docs/apim/4.8/self-hosted-installation-guides/kubernetes/vanilla-kubernetes.md
```diff
@@ -730,7 +730,7 @@ Keep the tunnel command running in a separate terminal window. The tunnel must r
     # Ensure you have PostgreSQL, Redis, and Elasticsearch services running in your cluster
     ```
 
-
+> If your Kubernetes cluster does not support IPV6 networking, both the ui and portal deployments need to define the `IPV4_ONLY` environment variable with `"true"` as a value.
```
Lyd1aCla1r3, line 733:
```suggestion
> If your Kubernetes cluster does not support IPV6 networking, both the UI and Portal deployments must set the `IPV4_ONLY` environment variable to `true`.
```

Lyd1aCla1r3 (approved):

PR #1224, opened 2026-03-08
AGENT-DRAFT: GKO-2006 — Client Certificate Management for Applications

Lyd1aCla1r3:
Automated update: gko-2006

Product: gko
Version: 4.11


benoitgravitee:
Hey,
A few things
* We are going to come up with a UI for this feature. Maybe the whole APIM chapter on the API endpoints it not really useful, although not false.
* For the GKO part, it is too light. The provided `.md` file in GKO-2006 is not really reflected on the result. There no examples examples on how manage, reference or rotate certificate.
* Status is rather internal although returned by the API, not sure it is of use for the end user

PR #1549, opened 2026-05-18
test vale component file utils

tolaore:


tolaore:
confirmation this works now

PR #1567, opened 2026-05-21
test(vale): add deliberate-error file to exercise Vale workflow

tolaore:
Covers every docs.Spelling resolution path (curated map, pyspellchecker, accept.txt, no-candidate) plus existing Vale rules (Vale.Terms, Microsoft.Contractions, Microsoft.We, Microsoft.Adverbs, Vale.Repetition). Delete this file after the workflow has been verified end-to-end.


tolaore:
this now works as well, as the docs.spelling works with the suggestino box

PR #1605, opened 2026-06-01
Updates to Gateway distributed sync docs

garethbrinn:


Lyd1aCla1r3:
@garethbrinn please use this as a test case for your skills :) Vale doesn't have sufficient scope to catch some of the errors, so it would be good to run this through your style guide mds a few times to record the deltas.

mukul-tyagi08 (approved):

tolaore (approved):

PR #1658, opened 2026-06-03
AGENT-DRAFT: APIM-13666 — Kafka Virtual Cluster & Multi-Cluster Routing [code+dev-prs]

Lyd1aCla1r3:
Automated update: apim-13666

Product: apim
Version: 4.12

### DIFF
--- a/docs/apim/4.12/kafka-gateway/create-and-configure-kafka-apis/create-kafka-apis.md
+++ b/docs/apim/4.12/kafka-gateway/create-and-configure-kafka-apis/create-kafka-apis.md
@@ -54,15 +54,15 @@ The only field to configure for the Kafka API entrypoint is the **Host prefix**,
 
 Select the endpoint type for your Kafka API. The available endpoint types are:
 
-- **Standalone**: Configure a direct connection to a single Kafka cluster by specifying bootstrap servers and security settings.
-- **Clusters**: Select a deployed cluster from the cluster registry.
-- **Virtual Clusters**: Select a deployed virtual cluster that aggregates multiple backend clusters.
+- **Broker** (connector `native-kafka`): Configure a direct connection to a single Kafka cluster by specifying bootstrap servers and security settings inline on the API.
+- **Cluster** (connector `native-kafka-cluster`): Reference a deployed Kafka cluster from the cluster registry. Only clusters in `DEPLOYED` state are listed.
+- **Virtual Cluster** (connector `native-kafka-virtual-cluster`): Reference a deployed virtual cluster that aggregates multiple backend clusters. Only virtual clusters in `DEPLOYED` state are listed.
 
 <figure><img src="../../.gitbook/assets/endpoint-create-kafka.png" alt=""><figcaption></figcaption></figure>
 
-### Standalone Endpoint Configuration
+### Broker Endpoint Configuration
 
-When you select **Standalone**, configure the bootstrap server list and security protocol to talk to the cluster directly.
+When you select **Broker**, configure the bootstrap server list and security protocol to talk to the cluster directly.
 
 Of the following configuration settings, only entering a host/port pair is required. Modifying any other configuration parameters is optional.
 
@@ -133,7 +133,7 @@ The Gravitee plans supported by Kafka APIs are summarized below, in increasing o
 <table><thead><tr><th width="201">Plan</th><th>Description</th></tr></thead><tbody><tr><td>Keyless (public)</td><td>When configured, this plan does not add security. It is considered an "open" plan.</td></tr><tr><td>API Key</td><td>The gateway only accepts connections from clients that pass an API key corresponding to a valid subscription to the proxy in the client properties. The API key is used as the password, and the md5 hash of the API key is used as the username, as part of the SASL/SSL with SASL PLAIN authentication method.</td></tr><tr><td>JWT</td><td>The gateway only accepts connections from clients that pass a valid JWT with a client ID claim corresponding to a valid subscription to the proxy in the client properties. This is equivalent to SASL/SSL with SASL OAUTHBEARER authentication, where the JWT is used as the OAuth token.</td></tr><tr><td>OAuth2</td><td>The gateway only accepts connections from clients that pass a valid OAuth token with a client ID corresponding to a valid subscription to the proxy in the client properties. This is equivalent to SASL/SSL with SASL OAUTHBEARER authentication.</td></tr></tbody></table>
 
 {% hint style="info" %}
-To learn more about how plans function in Gravitee, refer to the [plans](configure-kafka-apis/mtls-plans.md).
+To learn more about how plans function in Gravitee, refer to the [plans](../../secure-and-expose-apis/plans/) documentation. For mTLS plan configuration with Kafka APIs, see [mTLS plans](configure-kafka-apis/mtls-plans.md).
 {% endhint %}
 
 Individual plan configurations as they pertain to Kafka APIs are described in detail below.
--- a/docs/apim/4.12/kafka-gateway/create-and-configure-kafka-apis/create-kafka-apis.md
+++ b/docs/apim/4.12/kafka-gateway/create-and-configure-kafka-apis/create-kafka-apis.md
@@ -52,9 +52,17 @@ The only field to configure for the Kafka API entrypoint is the **Host prefix**,
 
 ## Endpoints
 
-The Kafka endpoint configuration is the Kafka cluster being proxied by your API. Configure the bootstrap server list and security protocol to talk to the cluster.
+Select the endpoint type for your Kafka API. The available endpoint types are:
+
+- **Standalone**: Configure a direct connection to a single Kafka cluster by specifying bootstrap servers and security settings.
+- **Clusters**: Select a deployed cluster from the cluster registry.
+- **Virtual Clusters**: Select a deployed virtual cluster that aggregates multiple backend clusters.
 
 <figure><img src="../../.gitbook/assets/endpoint-create-kafka.png" alt=""><figcaption></figcaption></figure>
+
+### Standalone Endpoint Configuration
+
+When you select **Standalone**, configure the bootstrap server list and security protocol to talk to the cluster directly.
 
 Of the following configuration settings, only entering a host/port pair is required. Modifying any other configuration parameters is optional.
 
@@ -125,7 +133,7 @@ The Gravitee plans supported by Kafka APIs are summarized below, in increasing o
 <table><thead><tr><th width="201">Plan</th><th>Description</th></tr></thead><tbody><tr><td>Keyless (public)</td><td>When configured, this plan does not add security. It is considered an "open" plan.</td></tr><tr><td>API Key</td><td>The gateway only accepts connections from clients that pass an API key corresponding to a valid subscription to the proxy in the client properties. The API key is used as the password, and the md5 hash of the API key is used as the username, as part of the SASL/SSL with SASL PLAIN authentication method.</td></tr><tr><td>JWT</td><td>The gateway only accepts connections from clients that pass a valid JWT with a client ID claim corresponding to a valid subscription to the proxy in the client properties. This is equivalent to SASL/SSL with SASL OAUTHBEARER authentication, where the JWT is used as the OAuth token.</td></tr><tr><td>OAuth2</td><td>The gateway only accepts connections from clients that pass a valid OAuth token with a client ID corresponding to a valid subscription to the proxy in the client properties. This is equivalent to SASL/SSL with SASL OAUTHBEARER authentication.</td></tr></tbody></table>
 
 {% hint style="info" %}
-To learn more about how plans function in Gravitee, refer to the [plans](../../secure-and-expose-apis/plans/) documentation. For mTLS plan configuration with Kafka APIs, see [mTLS plans](configure-kafka-apis/mtls-plans.md).
+To learn more about how plans function in Gravitee, refer to the [plans](configure-kafka-apis/mtls-plans.md).
 {% endhint %}
 
 Individual plan configurations as they pertain to Kafka APIs are described in detail below.
--- a/docs/apim/4.12/kafka-gateway/create-and-configure-kafka-clusters.md
+++ b/docs/apim/4.12/kafka-gateway/create-and-configure-kafka-clusters.md
@@ -13,7 +13,13 @@ metaLinks:
 This feature is in tech preview. Contact your customer team to request access to this feature.
 {% endhint %}
 
-The Kafka UI is accessible from the APIM Console. It is the user interface from which you can create and manage Kafka clusters, configure cluster connection information, and manage user access and permissions.
+The Kafka UI is accessible from the APIM Console. It is the user interface from which you can create and manage Kafka clusters and virtual clusters, configure cluster connection information, manage cluster lifecycle (deploy, undeploy, delete), and manage user access and permissions.
+
+The console navigation includes a **Kafka** menu with three sub-items:
+
+* **Standalone**
+* **Clusters**
+* **Virtual Clusters**
 
 ## Prerequisites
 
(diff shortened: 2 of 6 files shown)


docs/apim/4.12/kafka-gateway/create-and-configure-kafka-clusters.md
```diff
@@ -125,3 +187,38 @@ To add members to your Kafka cluster, complete the following steps:
...
+
+## Restrictions
+
+* Only the `PLAIN` SASL mechanism supports credential replay for cross-cluster operations. `AWS_MSK_IAM`, `SCRAM-SHA-*`, and `GSSAPI` mechanisms are not captured and cannot be replayed on backend connections.
+* Producer ID sessions are lost on gateway pod restart or round-robin reconnection unless a distributed `CacheManager` (Hazelcast or Redis) is configured. The gateway returns `PRODUCER_FENCED` (error code 90) instead of `UNKNOWN_PRODUCER_ID` (error code 59) when a producer ID session is missing. This forces Kafka clients to re-initialize the producer.
+* Virtual clusters strip KIP-848 and KIP-932 group APIs (consumer group protocol v3) from `ApiVersions` responses. Clients downgrade to the classic group protocol.
+* Backend connection pooling is scoped per-API, not process-wide. Multiple APIs targeting the same backend cluster open separate connection pools.
+* Virtual cluster metadata (controller ID, cluster ID) is cached after the first metadata merge. Until then, frame rewriters fall back to per-cluster values, which may cause instability across client reconnects.
+
+## Database schema changes
```
ThibaudAV: I'm wondering why we're talking about the database schema, IMHO internal detail , but I don't have the full picture, so maybe it's relevant 🤷‍♂️

docs/apim/4.12/kafka-gateway/create-and-configure-kafka-clusters.md
```diff
@@ -125,3 +187,38 @@ To add members to your Kafka cluster, complete the following steps:
...
+
+The API creation flow for Native Kafka APIs requires explicit endpoint selection in Step 3 (Endpoints). The previous auto-selection of both entrypoint and endpoint has been removed.
+
+The `KAFKA_CLUSTER_CONNECTION` endpoint type has been renamed to `KAFKA_CLUSTER_STANDALONE` with the UI label "Standalone".
+
+## Restrictions
+
+* Only the `PLAIN` SASL mechanism supports credential replay for cross-cluster operations. `AWS_MSK_IAM`, `SCRAM-SHA-*`, and `GSSAPI` mechanisms are not captured and cannot be replayed on backend connections.
+* Producer ID sessions are lost on gateway pod restart or round-robin reconnection unless a distributed `CacheManager` (Hazelcast or Redis) is configured. The gateway returns `PRODUCER_FENCED` (error code 90) instead of `UNKNOWN_PRODUCER_ID` (error code 59) when a producer ID session is missing. This forces Kafka clients to re-initialize the producer.
+* Virtual clusters strip KIP-848 and KIP-932 group APIs (consumer group protocol v3) from `ApiVersions` responses. Clients downgrade to the classic group protocol.
```
ThibaudAV: KIP-848 i's working. no downgrade 
KIP-932 it's share groups feature it's juste not supported

ThibaudAV: -> `"Virtual clusters strip the share-group API (KIP-932) from ApiVersions responses; share groups are not supported`

ThibaudAV: And I don't understand why we're discussing restriction in the context of clusters here, because it's a limitation of the virtual cluster. But since I don't fully understand the documentation, I might be mistaken.

docs/apim/4.12/kafka-gateway/create-and-configure-kafka-clusters.md
```diff
@@ -125,3 +187,38 @@ To add members to your Kafka cluster, complete the following steps:
...
+## API creation flow
+
+The API creation flow for Native Kafka APIs requires explicit endpoint selection in Step 3 (Endpoints). The previous auto-selection of both entrypoint and endpoint has been removed.
+
+The `KAFKA_CLUSTER_CONNECTION` endpoint type has been renamed to `KAFKA_CLUSTER_STANDALONE` with the UI label "Standalone".
+
+## Restrictions
+
+* Only the `PLAIN` SASL mechanism supports credential replay for cross-cluster operations. `AWS_MSK_IAM`, `SCRAM-SHA-*`, and `GSSAPI` mechanisms are not captured and cannot be replayed on backend connections.
+* Producer ID sessions are lost on gateway pod restart or round-robin reconnection unless a distributed `CacheManager` (Hazelcast or Redis) is configured. The gateway returns `PRODUCER_FENCED` (error code 90) instead of `UNKNOWN_PRODUCER_ID` (error code 59) when a producer ID session is missing. This forces Kafka clients to re-initialize the producer.
```
ThibaudAV: the PRODUCER_FENCED (90) vs UNKNOWN_PRODUCER_ID (59) behaviour is correct (confirmed in ProduceRequestPidRewriter), but this bullet misses the more important user-facing distinction: idempotent producers are supported on MESH (via MultiClusterInitProducerIdHandler + ProduceRequestPidRewriter), while transactional producers are not. Please split this into two bullets so users know which producer mode works.

Suggested:

> - Idempotent producers (enable.idempotence=true) are supported on virtual clusters. Producer ID sessions are kept in-memory by default; configure a distributed CacheManager (Hazelcast or Redis) to preserve them across gateway pod restarts or round-robin reconnections. When a session is missing, the gateway replies PRODUCER_FENCED (error code 90) instead of UNKNOWN_PRODUCER_ID (59), which forces the client to re-initialize the producer.
> - Transactional producers are not supported on virtual clusters.

In my opinion, other than mentioning the transactional limitation, there's no need to bring up the other one. It's supported, so the customer doesn't need to be concerned about it. We don't list every supported feature, otherwise we'd get lost in the details.

docs/apim/4.12/kafka-gateway/create-and-configure-kafka-clusters.md
```diff
@@ -80,6 +140,8 @@ In the **Configuration** tab, you can configure the following elements of the cl
 
       <figure><img src="../.gitbook/assets/9FAEF4B2-47B4-4D19-B2D2-D063ED96CAED_1_201_a.jpeg" alt=""><figcaption></figcaption></figure>
 
+The cluster **Configuration** tab uses custom autocomplete controls (`cluster-type`, `cluster-connection-type`) that filter to deployed clusters and validate against the deployed clusters API.
```
ThibaudAV: The mention of `cluster-type` / `cluster-connection-type` is an Angular form-control identifier — internal UI detail with no user value. The only user-facing fact is that the dropdown filters to deployed clusters. Suggest replacing this paragraph with:

> The cluster autocomplete only lists clusters in `DEPLOYED` state; undeployed and pending clusters are hidden.

docs/apim/4.12/kafka-gateway/create-and-configure-kafka-clusters.md
```diff
@@ -125,3 +187,38 @@ To add members to your Kafka cluster, complete the following steps:
...
+
+* `DEPLOYED` (green badge)
+* `PENDING` (yellow badge)
+* `UNDEPLOYED` (gray badge)
+
+## API creation flow
+
+The API creation flow for Native Kafka APIs requires explicit endpoint selection in Step 3 (Endpoints). The previous auto-selection of both entrypoint and endpoint has been removed.
+
+The `KAFKA_CLUSTER_CONNECTION` endpoint type has been renamed to `KAFKA_CLUSTER_STANDALONE` with the UI label "Standalone".
```
ThibaudAV: The "Cluster list tables" and "API creation flow" sections read like release-note bullets rather than how-to content:

- The Status badge colours are obvious from the UI and do not need to be enumerated.
- "the previous auto-selection of both entrypoint and endpoint has been removed" is changelog content; users coming to this page do not know what the previous behaviour was.
- The internal type rename `KAFKA_CLUSTER_CONNECTION` → `KAFKA_CLUSTER_STANDALONE` is invisible to users (they only see the UI label "Standalone").

Suggest removing both sections from this page. Anything worth keeping (e.g. that endpoint selection is now an explicit step) can be folded into the existing "Create Kafka APIs" walkthrough.

docs/apim/4.12/kafka-gateway/kafka-virtual-cluster-gateway-configuration-reference.md
```diff
@@ -0,0 +1,16 @@
+# Kafka Virtual Cluster Gateway Configuration Reference
+
+## Gateway Configuration
+
+### Routing Mode
+
+| Property | Description | Example |
+|:---------|:------------|:--------|
+| `gravitee.services.kafka.routing.mode` | Kafka routing mode. Determines how the gateway selects backend brokers. Valid values: `HOST` (route by hostname), `PORT` (route by port). | `HOST` |
```
ThibaudAV: ⚠️ This looks like a hallucinated property !!!
`gravitee.services.kafka.routing.mode` does not exist anywhere in `gravitee-reactor-native-kafka` (grepped the whole repo). Please double-check every property in this new reference file against the actual code; the rest of the table is suspect for the same reason.

The actual gateway property is `kafka.routingMode` (see `KafkaRoutingMode.PROPERTY_KEY`). There is no `gravitee.services.*` namespace and no dot between `routing` and `mode`. In `gravitee.yml`:

```yaml
kafka:
  routingMode: host   # or "port"
```

Values are parsed case-insensitively and default to `host`.

This is also the only switch in the table, but HOST mode has several companion properties that users actually need to configure with their wildcard TLS certificate. Please add them (all confirmed in `DefaultKafkaAcceptor` / `DefaultBrokerResolver`):

- `kafka.routingHostMode.defaultDomain` — wildcard suffix (e.g. `kafka.local`)
- `kafka.routingHostMode.defaultPort` — listener port (default `9092`)
- `kafka.routingHostMode.brokerPrefix` — prefix used to build per-broker hostnames (default `broker-`)
- `kafka.routingHostMode.domainSeparator` — separator between the broker id and the API host (default `-`)
- `kafka.routingHostMode.bootstrapDomainPattern` — bootstrap hostname pattern
- `kafka.routingHostMode.brokerDomainPattern` — per-broker hostname pattern

docs/apim/4.12/kafka-gateway/kafka-virtual-cluster-gateway-configuration-reference.md
```diff
@@ -0,0 +1,16 @@
...
+
+### Endpoint Attributes
+
+| Property | Description | Example |
+|:---------|:------------|:--------|
+| `gravitee.attributes.endpoint.nativekafkacluster` | Attribute prefix for native Kafka cluster endpoint configuration. | N/A |
```
ThibaudAV: Same hallucination pattern as the routing-mode entry — these are **not** gateway configuration properties. `gravitee.attributes.endpoint.nativekafkacluster` and `gravitee.attributes.endpoint.nativekafkavirtualcluster` are the `ATTRIBUTE_PREFIX` Java constants in `NativeKafkaClusterEndpointConnectorConfiguration` and `NativeKafkaVirtualClusterEndpointConnectorConfiguration`. They are used internally as keys for context-bound EL attributes on a request; a user setting them in `gravitee.yml` will get no effect.

Please remove this section. If there is a genuine need to document internal attributes (for EL authors), put it in a dedicated internals/EL reference page, not in a gateway configuration reference.

docs/apim/4.12/kafka-gateway/create-and-configure-kafka-apis/create-kafka-apis.md
```diff
@@ -52,10 +52,18 @@ The only field to configure for the Kafka API entrypoint is the **Host prefix**,
...
+
+- **Standalone**: Configure a direct connection to a single Kafka cluster by specifying bootstrap servers and security settings.
+- **Clusters**: Select a deployed cluster from the cluster registry.
+- **Virtual Clusters**: Select a deployed virtual cluster that aggregates multiple backend clusters.
 
 <figure><img src="../../.gitbook/assets/endpoint-create-kafka.png" alt=""><figcaption></figcaption></figure>
 
+### Standalone Endpoint Configuration
+
+When you select **Standalone**, configure the bootstrap server list and security protocol to talk to the cluster directly.
```
ThibaudAV: The endpoint-type labels are wrong. This page documents the **API creation** flow (Step 3 - Endpoints), where the labels come from the endpoint connector plugins:

| Connector id | Label shown in API creation |
|---|---|
| `native-kafka` | **Broker** |
| `native-kafka-cluster` | **Cluster** |
| `native-kafka-virtual-cluster` | **Virtual Cluster** |



"Standalone / Clusters / Virtual Clusters" are the labels in the **Cluster admin sidebar** (`gravitee-apim-console-webui/.../clusters/...`), which is a different UI used to manage deployed cluster registry entries. The PR mixes the two contexts.

Suggested rewrite:
> Select the endpoint type for your Kafka API:
> - **Broker** (connector `native-kafka`): configure a direct connection to a single Kafka cluster by specifying bootstrap servers and security settings inline on the API.
> - **Cluster** (connector `native-kafka-cluster`): reference a deployed Kafka cluster from the cluster registry. Only clusters in `DEPLOYED` state are listed.
> - **Virtual Cluster** (connector `native-kafka-virtual-cluster`): reference a deployed virtual cluster that aggregates multiple backend clusters. Only virtual clusters in `DEPLOYED` state are listed.

Also rename the "Standalone Endpoint Configuration" subsection on line 63 to "Broker Endpoint Configuration".

docs/apim/4.12/kafka-gateway/create-and-configure-kafka-apis/create-kafka-apis.md
```diff
@@ -125,7 +133,7 @@ The Gravitee plans supported by Kafka APIs are summarized below, in increasing o
...
 
 {% hint style="info" %}
-To learn more about how plans function in Gravitee, refer to the [plans](../../secure-and-expose-apis/plans/) documentation. For mTLS plan configuration with Kafka APIs, see [mTLS plans](configure-kafka-apis/mtls-plans.md).
+To learn more about how plans function in Gravitee, refer to the [plans](configure-kafka-apis/mtls-plans.md).
```
ThibaudAV: This edit broke the general plans link. The original line linked the word "plans" to `../../secure-and-expose-apis/plans/` (the general plans documentation) and had a separate sentence about mTLS. The current version points "plans" at the mTLS-specific page, which is not the same content. Please restore:

> To learn more about how plans function in Gravitee, refer to the [plans](../../secure-and-expose-apis/plans/) documentation. For mTLS plan configuration with Kafka APIs, see [mTLS plans](configure-kafka-apis/mtls-plans.md).

docs/apim/4.12/release-information/release-notes/apim-4.12.md
```diff
@@ -59,6 +59,16 @@
 * The `service.name` must match between tracer and logger (default: `gio_apim_gateway`) for Grafana correlation to function correctly.
 <!-- /PIPELINE:APIM-13463 -->
 
+
+<!-- PIPELINE:APIM-13666 -->
+#### **Kafka Virtual Cluster and Multi-Cluster Routing**
+
+* Aggregate multiple Kafka clusters behind a single virtual endpoint to simplify multi-region or multi-tenant deployments.
+* Route client requests transparently across backend clusters based on topic, partition, or consumer group affinity while merging metadata into a unified view.
+* Maintain producer ID session mappings to ensure transactional producer stability and replay SASL PLAIN credentials for cross-cluster operations.
```
ThibaudAV: Transactional producers are **not** supported on MESH virtual clusters. The PID session machinery preserves **idempotent** producer state, which is a different guarantee. Please reword:

> Maintain producer ID session mappings to preserve **idempotent** producer state across multi-cluster routing; replay SASL PLAIN credentials for cross-cluster operations. Transactional producers are not yet supported on virtual clusters.

And same maybe we can ignore idempotent support detail

ThibaudAV:
Beyond the inline comments, here are MESH-specific topics I think the doc would benefit from covering. None of them are blocking, but several would save users from confusing first encounters.

---

**1. Multi-pod prerequisite — distributed cache** *(critical for production)*

When you run more than one gateway pod behind a load balancer, virtual cluster routing state must be shared across pods. Configure a distributed cache backend in `gravitee.yml`:

```yaml
cache:
  type: hazelcast   # or "redis"
```

Without a distributed cache, the gateway falls back to local in-memory storage per pod. The following stops working correctly:

- **Idempotent producers** — a client whose connection lands on a different pod after a reconnect loses its producer ID session and receives `PRODUCER_FENCED`, forcing a producer re-initialization.
- **Consumer group routing** — each pod maintains its own view of which backend hosts a group, which can lead to inconsistent routing.

A single-pod gateway works without a distributed cache. Hazelcast and Redis configuration are documented in the Gravitee Node Cache module README.

---

**2. Topic creation contract on a virtual cluster**

Suggested addition near the virtual cluster section:

> `CreateTopics` requests sent through a virtual cluster are created on the first backend cluster listed in the virtual cluster's configuration. If that backend is unavailable, the gateway falls back to the next configured backend, and so on. Routing topics to a specific backend (e.g. by prefix) is not yet supported — to place a topic on a specific backend, create it directly on that backend.

---

**3. ACL operations**

Suggested addition under Restrictions:

> ACL operations (`CreateAcls`, `DeleteAcls`, `DescribeAcls`) are not supported through a virtual cluster — requests are rejected with `SECURITY_DISABLED` and AdminClient surfaces a `SecurityDisabledException`. Manage ACLs directly on each backend Kafka cluster.

---

**4. Virtual broker IDs (TLS planning)**

Virtual clusters rewrite backend broker IDs to keep them unique across the aggregated view:

```
virtualBrokerId = (backendIndex + 1) * 10000 + realBrokerId
```

where `backendIndex` is the zero-based position of the backend in the virtual cluster's `backends` array. Broker `3` on the first backend becomes virtual broker `10003`; broker `0` on the second backend becomes `20000`.

**TLS implications.** In HOST routing mode, each broker is reached at `broker-<virtualBrokerId>-<apiHost>.<domain>`. With a wildcard certificate (`*.<domain>`), nothing extra is required. Without a wildcard, you must anticipate every per-broker hostname your virtual cluster will surface (one per real broker across all backends, using the virtual IDs above) and issue an SAN-covered certificate for each. Missing a hostname will surface as a TLS handshake failure on client reconnect.

---

**5. Shadow consumer groups (short note)**

Suggested hint somewhere in the Virtual Cluster section:

> {% hint style="info" %}
> When a client consumes from a virtual cluster, the gateway creates one internal consumer group per backend cluster, named `<groupId>__shadow-c<N>` (indexed from zero). These groups are managed by the gateway — do not delete them manually. Inspect consumer groups through the virtual cluster's bootstrap address; the gateway re-aggregates the shadow groups into the single client-visible group.
> {% endhint %}

---

**6. Pointer to the runnable quick-setup** *(idea only — not sure if conventional in your docs)*

The `gravitee-api-management` repository ships a runnable example under [`docker/quick-setup/native-kafka-mesh/`](https://github.com/gravitee-io/gravitee-api-management/tree/master/docker/quick-setup/native-kafka-mesh). It launches two single-node KRaft brokers (EU + US), the gateway, and the management stack; the README walks through registering the backends, creating a virtual cluster, deploying an API with a Keyless plan, producing on each backend, consuming the merged view through the virtual cluster, and inspecting the resulting consumer-group state.

(The example lives under `master` once [PR #17375](https://github.com/gravitee-io/gravitee-api-management/pull/17375) is merged.)

PR #1810, opened 2026-06-23
AGENT-DRAFT: APIM-14501 — Token-Bucket Rate Limiting [code+dev-prs]

Lyd1aCla1r3:
Automated update: apim-14501

Product: apim
Version: 4.12


wbabyte (approved):

wbabyte:
**Message-API metering unit is unspecified — should state "per connection (Initial connection), not per message"**

The doc covers message APIs but never states the *metering unit* for them. The only unit sentence is HTTP-centric — *"Each request consumes one token… (or message stream interruption on message APIs)"* — and the message section only describes the rejection mechanism (`interruptMessagesWith`). A reader will reasonably assume **one token per message**, which is wrong.

Verified locally on a built 4.12.x gateway (policy `5.0.0-alpha.5`, MongoDB rate-limit store): on a V4 message API the policy meters at the **Initial connection** phase — **one token per connection**, regardless of how many messages flow on it:

- A single `http-get` subscribe connection held open ~3s delivered **29 messages** (`"content":"hello"` ×29) but consumed **exactly one token** (`X-Rate-Limit-Remaining` 5 → 4). Per-message metering would have burned 29.
- Burst is enforced per connection: 14 rapid publish connections → **10×202 then 4×429** (burst capacity 10), then refill at the configured rate.

Suggested addition to the message-API section:

> On V4 message APIs the policy meters at the **Initial connection** phase — **one token per connection** (publish or subscribe), not per message. A single connection can stream any number of messages on that one token; when the bucket is empty the message stream is interrupted instead of returning HTTP 429.

This matches the plugin README ("one consume per phase") and the implemented behaviour in APIM-14497, and it also tells Console users where to attach the policy in Policy Studio (the **Initial connection** point, not the per-message point).

tolaore (approved):

PR #2056, opened 2026-07-28
[AGENT-DRAFT] docs(apim): add LangChain guide for the LLM proxy

garethbrinn:
Drafted by the **Field Content Pipeline** from a field-authored integration guide (*Securing LangChain agents with Gravitee*, Sam Prodger, 2026-07-28), verified against APIM 4.12.4 by its author.

⚠️ **This is an agent draft.** It requires review per `documentation-codebase/backlog-agent/agent-draft-review.md` and must not merge until the questions below are answered.

---

## Part 1 — Questions for the author (@Sam Prodger)

Ordered by consequence. Each has the evidence attached.

### 1. `pii-filtering` is **not** request-phase only — was response masking ever enabled when you tested?

The guide states *"Request phase only: it never masks what comes back from the model"*, and lists it under Limits.

The source disagrees. `PiiFilteringPolicy.onRequest()` registers a response-phase hook at runtime:

```java
if (!configuration.skipResponsePayloadFiltering()) {
    ctx.addActionOnResponse(this, c -> { /* redacts the response body */ });
}
```

`schema-form.json` gives `skipResponsePayloadFiltering` a default of `false`, so **response redaction is on by default**. The guide's own config sets it to `true`, which is exactly what disables it.

Your observation that the manifest declares `REQUEST` on every entrypoint is **correct** — `http_proxy=REQUEST`, `llm_proxy=REQUEST`, `mcp_proxy=REQUEST`. That's the phase the policy is *attached* in, not the phases it acts in.

**Downstream impact:** the companion Databricks guide builds a custom Groovy `scope: RESPONSE` masking policy on this premise. If the shipped policy already covers it, that section should be removed rather than published. Please confirm before we process that asset.

### 2. Should the `GatewayChatModel` implementation be published at all?

The custom `BaseChatModel` subclass is ~60 lines of LangChain integration code that runs in your agent process. `critical-rules.md` rule 7 forbids documenting third-party products, and your own header notes *"No custom gateway plugins are used or required."* It's cut from this draft in favour of the one-line `ChatOpenAI` path.

The trade-off: the retry/tracing/null-message handling in it is genuinely useful and is now nowhere. Would you rather it lived in a field-owned repo or blog we can link to?

### 3. Accurate-but-incomplete: a monetary budget policy does exist

The guide says `token-ratelimit` *"does not price anything… it is not a monetary budget"*, which is true of that policy. But `gravitee-policy-token-ratelimit` also ships **`cost-ratelimit`** — free, `llm_proxy=REQUEST`:

> Limits how much money a consumer may spend on model calls over a period of time.

It reads `llmproxy.usage.sent.cost` / `received.cost` derived from the same per-model `inputPrice`/`outputPrice` you already configure, budgets in micro-dollars, and supports `ADAPTIVE` reservation. It even documents releasing a reservation on a semantic-cache hit.

This sits directly against the guide's framing that *"nothing caps token spend… you find out on the invoice"*. Want it covered?

### 4. Claims left unverified — no live environment was provided

Step 5 didn't run, so these keep your original verification status and are excluded from the page:

- Failover behaviour and the response `model` field naming the fallback (you flagged this as verified live once, not on demand).
- The "HTTP 200 with `message: null` under load" provider behaviour.
- The `929ms` → `12ms` semantic cache timings.

### 5. Observed internals — one resolved in your favour, one still open

- **Resolved:** the `llm-proxy` metric field names are **already documented** in `analyze-and-monitor-apis/reporters/README.md`. They're public contract, so your *"not publicly documented, do not build a contract on them"* caveat is unnecessary for those.
- **Still open:** the `semantic-caching_cache-hit` / `cache-hit-score` / `cache-hit-tokens-saved` names appear nowhere in the docs. Publishing them would create a contract. Recommend withholding.

### 6. Enterprise pack mapping unconfirmed

All four licence claims in your guide are **correct** — verified at two layers (`feature=` in `plugin.properties`, plus no `LicenseManager` gating in any policy source):

| Policy | `feature=` | Enterprise? | Your claim |
|---|---|---|---|
| `pii-filtering` | `apim-policy-pii-filtering` | Yes | ✅ |
| `ai-semantic-caching` | `apim-ai-policy-semantic-cache` | Yes | ✅ |
| `ai-prompt-guard-rails` | *(none)* | No | ✅ |
| `token-ratelimit` | *(none)* | No | ✅ |

What I couldn't confirm is the **pack**. `enterprise-edition.md` doesn't map either feature name to a named pack. The 4.13 `pii-filtering` page claims an `agent-mesh` pack. Do you know which pack a customer actually buys?

---

**Pages:** 1 new — `docs/apim/4.13/ai-agent-management/llm-proxy/consume-your-llm-proxy-with-langchain.md`, plus `SUMMARY.md` and the section `README.md` card table. Placed as a sibling to the existing OpenAI SDK and Claude Code guides, which are the established precedent for this page shape.

**Version:** authored against 4.12.4, published into **4.13**. All shipped claims were re-verified against 4.13.

**Verdicts:** 6 confirmed and shipped · 1 contradicted · 3 accurate-but-incomplete or undetermined · ~14 cut as already documented · 9 cut as third-party.

**Confirmed and shipped:** the six-provider enum matches the 4.13 overview exactly · aliases resolve through the model registry · the prefix policy must permit unprefixed aliases · `X-Gravitee-Api-Key` is the documented header convention for API key plans.

**Cut as duplicates (rule 25):** route creation, endpoint config, failover config, and the guard-rails / token-rate-limit / PII / semantic-caching policy configuration are all already documented. The page links to them rather than restating.

**Cut as third-party (rule 7):** LangChain client internals, Gemini and Databricks endpoint shapes, `Retry-After` semantics, provider-side failure modes.

**Added, not in the source asset:** the Enterprise License prerequisite. The LLM proxy itself requires one and every sibling page states it — the guide didn't.

### Docs bugs found on the way (separate from this PR)

1. **`pii-filtering.md` streaming section looks stale.** It says a streaming request with response filtering enabled is rejected with a 400. The schema says SSE responses are *"filtered in place"* and the policy ships an `SseRedactingTransformer`.
2. **Inconsistent metric field names.** `reporters/README.md` uses `double_llm-proxy_sent-cost` (hyphen), `datadog-reporter.md` uses `double_llm-proxy_sent_cost` (underscore). At most one is right.
3. **Typos in the sibling OpenAI SDK page** — "your API K.", "yourt GAteway URL".


garethbrinn:
**Correction from a second pipeline run** (on the companion Databricks asset), plus one finding that affects this PR.

### Correction: I cut failover as "already documented" — too glib

Failover *is* documented at `create-and-configure-apis/configure-v4-apis/load-balancing-and-failover.md`, with all defaults. But that page scopes itself to "v4 **HTTP proxy** APIs", explicitly excludes Kafka, and mentions LLM zero times.

Both field guides configure failover on an `LLM_PROXY` and depend on it rotating providers. Whether the documented feature applies to that API type is **undetermined**, and it should have been question 7 here rather than a silent cut. Raised as question 4 on #2059.

### New: `OPEN_AI` does honour a custom target URL

Not a claim in this asset, but it is in the companion, and it may affect how the LangChain guide's endpoint config is read. `LLMProxyEndpointConnectorFactory` routes both providers identically:

```java
case OPEN_AI, OPEN_AI_COMPATIBLE -> new OpenAIResolver(config.getTarget());
```

Verified at the connector layer only — a console or schema difference could still explain the original observation.

### Question 1 here now has source-level confirmation

`OpenAIResolver` confirms the connector appends the consumer's API path to the configured target, so the "no `/v1`" guidance is right for the right reason. That's now documented in #2059.

One answer to question 1 here covers question 2 on #2059 — they're the same `pii-filtering` finding.

samprodger:
Good flags — re-reviewed the code and the brain dumps, agree. Live retest confirms it, both pieces corrected.

samprodger:
Follow-up on the pii-filtering question (full details on #2059): ran the live retest, confirmed. `pii-filtering` in the request flow with `skipResponsePayloadFiltering` at its default `false` masks the response too — proved it live, PII came back `[REDACTED]` on a route matching this guide's shape. My original "request phase only" claim was wrong, caused by my own test config having that flag set `true`. Correcting the LangChain guide's Limits section to match before it publishes.

samprodger (approved):
Content checks out — ChatOpenAI base_url/header pattern matches the verified X-Gravitee-Api-Key convention, model-alias/prefix guidance is correct, PII/failover claims already resolved in the thread above. Good to merge.

tolaore (approved):

PR #2059, opened 2026-07-28
[AGENT-DRAFT] docs(apim): document target URL for OpenAI-compatible LLM providers

garethbrinn:
Drafted by the **Field Content Pipeline** from *Securing Databricks with Gravitee*, a field-authored integration guide verified by its author against APIM 4.12.4.

⚠️ **Agent draft.** Requires review per `documentation-codebase/backlog-agent/agent-draft-review.md`.

**Note on scope:** the source asset covered four Databricks surfaces and ran ~500 lines. This PR is one small page. That is the expected outcome, explained in Part 2 — the great majority of the asset is either Databricks documentation the docs team can't publish, or Gravitee behaviour that's already documented.

---

### 1. `OPEN_AI` does honour a custom target URL

The guide says:

> The plain `OPEN_AI` provider does not honour a custom target URL, so keep `OPEN_AI_COMPATIBLE`.

`LLMProxyEndpointConnectorFactory` routes both through the same resolver with the same configured target:

```java
case OPEN_AI, OPEN_AI_COMPATIBLE -> new OpenAIResolver(config.getTarget());
```

Verified at the connector layer only. It's possible the console UI or the endpoint schema hides or ignores the target field for `OPEN_AI` — that would explain what you saw, and it would be a UI behaviour rather than a connector one. Do you recall whether the target was rejected, ignored at runtime, or simply not editable in the console?

### 2. The custom Groovy response-masking policy appears unnecessary

Your *"Response masking needs your own policy"* section rests on `pii-filtering` being request-phase only. It isn't — `PiiFilteringPolicy.onRequest()` registers a response hook via `ctx.addActionOnResponse(...)`, and `skipResponsePayloadFiltering` defaults to `false`.

Your manifest observation was correct. `REQUEST` on every entrypoint is the phase the policy is *attached* in, not the phases it acts in. Deploying it in a response flow does nothing because it belongs in the request flow, where it handles both.

**If confirmed, that whole section should be removed rather than published.** Same question as #1 on PR #2056 — one answer covers both.

### 3. The Groovy sandbox has a supported extension mechanism

Your *"The Groovy sandbox will fight you"* findings are real, but [groovy.md](https://documentation.gravitee.io/) documents a whitelist you can extend rather than work around, via `groovy.whitelist.mode: append` and `groovy.whitelist.list` in `gravitee.yml`, with the full list published.

Extending the whitelist is likely cleaner than the `HttpURLConnection` and `encodeBase64()` workarounds. Worth retesting before that guidance spreads to the Snowflake and Bedrock guides you have planned.

### 4. Does failover apply to `LLM_PROXY` APIs?

Both your guides configure `failover` on an `LLM_PROXY` and rely on it rotating providers. The [failover page](https://documentation.gravitee.io/) scopes itself to "v4 **HTTP proxy** APIs", explicitly excludes Kafka, and never mentions LLM.

Your settings are all non-default, which is fine and correctly flagged as choices in the LangChain guide. The open question is whether the documented feature is specified to apply to this API type at all. You flagged the 429 and 5xx conditions as never exercised live, so this is unverified on both sides.

### 5. Claims that stay unverified — no live environment

Runtime claims are excluded and remain at your original status: the once-verified Delta and Zerobus routes, the `929ms` → `12ms` cache timings, and the semantic-caching metric field names, which appear nowhere in the docs.

---

**Confirmed from source** (`gravitee-reactor-llm-proxy`):

- The provider enum is exactly six values, matching your list: `OPEN_AI`, `OPEN_AI_COMPATIBLE`, `GEMINI`, `BEDROCK`, `ANTHROPIC`, `VERTEX_AI`.
- The connector appends the consumer's API path to the configured target. `OpenAIResolver.resolveEndpointPath()` also strips a known API path if the target already carries one. This is the page's subject and the one piece of net-new content in the asset.

**Cut as third-party (rule 7)** — the docs team publishes the Gravitee side and links out for the rest:

Lakebase entirely, since the gateway doesn't front it and the material is Postgres and Databricks tuning advice · Unity Catalog DDL · Zerobus table-storage constraints, type encoding, and quotas · Statement Execution limits · all vendor GA and beta dates · the OAuth mint and its `authorization_details` grant.

**Cut as already documented (rule 25)** — the overlap map:

| Asset content | Existing page |
|---|---|
| LLM proxy creation and endpoint config | `proxy-your-llms.md` |
| Backend credential injection | `policy-reference/transform-headers.md` |
| Response transformation | `policy-reference/groovy.md` |
| Failover settings and defaults | `configure-v4-apis/load-balancing-and-failover.md` |
| Semantic caching, PII filtering, token rate limit | their policy reference pages |

**Placement is yours to decide.** This page could equally be a section of `proxy-your-llms.md`. It's drafted standalone because the pipeline no longer merges into existing pages on its own — closing it as already-covered is a legitimate outcome.

**Not included, deliberately:** the specific Databricks base URL. It's a Databricks fact under rule 7, and the page instead explains the rule the connector actually applies, which holds for any OpenAI-compatible provider.


samprodger:
Good flags — re-reviewed the code and the brain dumps, agree. Live retest confirms it, both pieces corrected.

samprodger:
Follow-up: ran the live retest. Deployed `pii-filtering` in the request flow with `skipResponsePayloadFiltering` left at its default `false` against a route matching `delta-query`'s shape (plain `http-proxy` entrypoint, real PII in the response). Confirmed: name, email and phone all came back `[REDACTED]`. Same request leaked them unmasked before, with the flag set `true`.

So this is fully closed now, not just a code-reading inference: the "request-phase only, response masking is your own job" claim was wrong, caused by my own test config. I've corrected the guide — the custom Groovy response-masking policy is no longer positioned as necessary, kept only as the documented option for masking a specific business column outside `pii-filtering`'s fixed PII category list. Ghost draft's updated. Thanks for catching this, it would have shipped wrong otherwise.

samprodger:
On Q1 (OPEN_AI custom target URL): not chasing this down further for now — the shipped page only documents OPEN_AI_COMPATIBLE and never makes a claim about plain OPEN_AI, so it doesn't block this content either way. Happy to verify the console behaviour separately if it's worth its own note later.

samprodger (approved):
Content checks out — target-URL/path-append behaviour matches the connector source, console steps are correct, and the page doesn't depend on the still-open OPEN_AI question (see comment above). Good to merge.

tolaore (approved):

PR #2066, opened 2026-07-28
[AGENT-DRAFT] Screen A2A agent traffic with the Guard Rails policy

samprodger:
Drafted by the **Field Content Pipeline** (author-run mode) from *The gateway that reads your agent's mail*, a field-authored piece on governing A2A agent traffic with Gravitee, verified by its author against a Gravitee 4.11 gateway. The asset was itself already unusually self-critical — most of its "Where this breaks" section is the author's own flagged gaps, not claims the pipeline had to correct.

⚠️ **Agent draft.** Requires review per `documentation-codebase/backlog-agent/agent-draft-review.md`.

---

## Part 1 — Questions for the author (@samprodger)

Reply in a comment, numbered to match. Prose is fine, no form.

### 1. Enterprise licence on `ai-prompt-guard-rails` — I can't confirm the pack

`gravitee-policy-ai-prompt-guard-rails/src/main/resources/plugin.properties` has no `feature=` line at all, which by rule 9 would mean it's not gated. But the existing published guide (`add-the-guard-rails-policy-to-your-llm-proxy.md`) lists an Enterprise Licence as a prerequisite anyway, and the new page follows that precedent rather than the plugin manifest. I don't have a way to settle this from the policy repo alone. Do you know whether this policy is gated at a level above the individual plugin, or is the existing guide's prerequisite list stale?

### 2. Two likely bugs in Gravitee's own published docs, found while verifying your claims

Not about your asset, about the docs it made me read closely:

* `ai-prompt-guard-rails.md`'s configuration table states the default `sensitivityThreshold` is `0.5`. The plugin's own `schema-form.json` says `0.8`.
* The same page states "Compatible API types: `PROXY`" only. The plugin manifest declares `proxy=REQUEST`, `llm_proxy=REQUEST`, **and** `a2a_proxy=REQUEST`.

Should I open these as separate docs-bug tickets, or is there a reason the published page is deliberately narrower than the manifest (e.g. `llm_proxy`/`a2a_proxy` support exists but isn't supported/announced yet)?

### 3. The "no method-level routing, no `skills` array" claim about `A2A_PROXY` — confirmed, but flagging per rule 29

I read every non-test `.java` file in `gravitee-reactor-a2a-proxy` (10 files total). There's no method dispatch on the JSON-RPC `method` field and no code that reads a card's `skills` array anywhere outside test fixtures. Your absence claim holds at that layer. Given rule 29's caution about absence claims, and that this speaks to product roadmap rather than current behaviour, flagging it as a question rather than a silent publish: comfortable with this shipping as stated, or would you rather it stayed softer ("not currently exposed") in case that's already in flight?

### 4. `promptLocation` needs `promptPreset: CUSTOM_PROMPT` — your guide doesn't say so, the page now does

Your write-up describes `promptLocation` as "the whole trick", one field. The schema disables that field unless `promptPreset` is set to `CUSTOM_PROMPT` (it's hidden under the default `ALL_PROMPTS`). The new page includes both fields. Worth fixing in your original guide too, so it doesn't propagate into the next A2A piece.

### 5. Claims cut as out of scope for this page, not wrong

The RFC 8693 token-exchange gap, the toxicity-classifier-not-enabled gap, and the JSON-RPC-method-blind-enforcement gap are all real limitations you flagged about *your build*, not about the Guard Rails policy this page documents. They didn't fit a page about pointing one policy at one field. Worth their own page or field note if you want them captured anywhere durable, they're good material.

---

**Source:** field-authored piece, Sam Prodger, verified against APIM 4.11. Self-declared verification status was unusually granular per claim (author explicitly separated "verified live" from "config says X but I never enabled it" throughout), which made Step 2 triage fast.

**Verdict table:**

| Claim | Bucket | Verdict | Source |
|---|---|---|---|
| `A2A_PROXY` rewrites agent card `url` and `supportedInterfaces[].url` to the gateway's own | A | Confirmed | `AgentCardProcessor.java` |
| `A2A_PROXY` only intercepts GET to `/.well-known/agent-card.json`, `/.well-known/agent.json`, `/extendedAgentCard` | A | Confirmed | `AgentCardProcessor.canHandle()` |
| `A2A_PROXY` has no method-level routing or `skills`-array handling | C | Confirmed absent, all non-test source read | `gravitee-reactor-a2a-proxy` (10 files) |
| `A2A_PROXY` is Enterprise, under a named pack | A | Confirmed, pack now named | `a2a-proxy-api-type.md`: "AI Agent Management pack" |
| Llama Prompt Guard 2, 22M param, ONNX, ships as a selectable model | A | Confirmed | `gravitee-policy-ai-prompt-guard-rails` README/docgen, cross-referenced HuggingFace model card |
| Model returns `BENIGN`/`MALICIOUS` labels | A | Confirmed | Matches published `ai-prompt-guard-rails.md` policy reference exactly |
| `promptLocation` accepts an EL expression, works against a JSON-RPC path | A | Confirmed | `schema-form.json`, existing policy reference EL example |
| `promptLocation` requires `promptPreset: CUSTOM_PROMPT` to be active | A | New finding, not in source asset | `schema-form.json` `disableIf` |
| `sensitivityThreshold` default is 0.8 | A | Contradicts published docs (which say 0.5) | `schema-form.json` vs `ai-prompt-guard-rails.md` |
| Guard Rails compatible with `PROXY` only | A | Contradicts published docs (manifest also declares `llm_proxy`, `a2a_proxy`) | `plugin.properties` vs `ai-prompt-guard-rails.md` |

**Cut as third-party (rule 7):** the RNLI stations/sea-conditions agent domain content, a2a-sdk implementation details, agent card `streaming` field self-declaration behaviour (real point, but about the a2a-sdk, not Gravitee).

**Cut as out of scope for a Guard Rails how-to (not wrong, just a different page):** RFC 8693 token exchange, the classifier-ordering-before-JSON-Threat-Protection lesson (folded into a short "Order this policy after payload validation" section instead, since it's genuinely Gravitee-configuration advice), the toxicity-classifier gap, JSON-RPC-method-blind enforcement.

**Overlap map:**

| Section of new page | Existing page | Overlap |
|---|---|---|
| A2A_PROXY licensing, architecture, what it does on a card fetch | `a2a-proxy-api-type.md` | Heavy — new page links out rather than restating |
| Guard Rails resource setup, model sizing, Alpine/ONNX gotcha | `ai-prompt-guard-rails.md`, `add-the-guard-rails-policy-to-your-llm-proxy.md` | Heavy — new page links out rather than restating |
| Pointing `promptLocation` at a JSON-RPC A2A envelope specifically | None found | None — this is the actual new content |

Heavy overlap on two of three sections. Flagging per Step 6: the genuinely new material here is narrow (the EL expression, the `promptPreset` prerequisite, the policy-ordering note). Placement and whether this justifies its own page versus a shorter addition to the existing Guard Rails reference page is the writer's call.

**Docs bugs found on the way:** the two `ai-prompt-guard-rails.md` discrepancies in Question 2 above.

### DIFF
--- a/docs/apim/4.13/SUMMARY.md
+++ b/docs/apim/4.13/SUMMARY.md
@@ -462,6 +462,7 @@
 * [AI Agent Management](ai-agent-management/README.md)
   * [Set Up an AI Agent (A2A) Proxy](ai-agent-management/create-an-a2a-proxy.md)
   * [Discover and Catalog AI Agents (A2A)](ai-agent-management/add-agents-to-your-agent-catalog.md)
+  * [Screen A2A agent traffic with the Guard Rails policy](ai-agent-management/screen-a2a-agent-traffic-with-the-guard-rails-policy.md)
   * [Convert REST APIs to an MCP Server](ai-agent-management/convert-your-apis-to-mcp-servers.md)
   * [LLM proxy](ai-agent-management/llm-proxy/README.md)
     * [Proxy your LLMs](ai-agent-management/llm-proxy/proxy-your-llms.md)
--- a/docs/apim/4.13/ai-agent-management/README.md
+++ b/docs/apim/4.13/ai-agent-management/README.md
@@ -6,7 +6,7 @@ description: An overview about AI Agent Management.
 
 
 
-<table data-view="cards"><thead><tr><th></th><th data-hidden data-card-target data-type="content-ref"></th></tr></thead><tbody><tr><td>Set up an AI Agent (A2A) Proxy</td><td><a href="create-an-a2a-proxy.md">create-an-a2a-proxy.md</a></td></tr><tr><td>Discover and Catalog AI Agents (A2A)</td><td><a href="add-agents-to-your-agent-catalog.md">add-agents-to-your-agent-catalog.md</a></td></tr><tr><td>Convert REST APIs to an MCP Server</td><td><a href="convert-your-apis-to-mcp-servers.md">convert-your-apis-to-mcp-servers.md</a></td></tr><tr><td>LLM Proxy</td><td><a href="llm-proxy/">llm-proxy</a></td></tr></tbody></table>
+<table data-view="cards"><thead><tr><th></th><th data-hidden data-card-target data-type="content-ref"></th></tr></thead><tbody><tr><td>Set up an AI Agent (A2A) Proxy</td><td><a href="create-an-a2a-proxy.md">create-an-a2a-proxy.md</a></td></tr><tr><td>Discover and Catalog AI Agents (A2A)</td><td><a href="add-agents-to-your-agent-catalog.md">add-agents-to-your-agent-catalog.md</a></td></tr><tr><td>Screen A2A agent traffic with the Guard Rails policy</td><td><a href="screen-a2a-agent-traffic-with-the-guard-rails-policy.md">screen-a2a-agent-traffic-with-the-guard-rails-policy.md</a></td></tr><tr><td>Convert REST APIs to an MCP Server</td><td><a href="convert-your-apis-to-mcp-servers.md">convert-your-apis-to-mcp-servers.md</a></td></tr><tr><td>LLM Proxy</td><td><a href="llm-proxy/">llm-proxy</a></td></tr></tbody></table>
 
 ## Overview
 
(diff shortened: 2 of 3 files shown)


samprodger:
Closing — this was a pipeline trial run, not intended to ship. Findings are still useful (see thread above) but this shouldn't be a live PR.

PR #2102, opened 2026-07-31
[AGENT-DRAFT] docs: layered governance for MCP tools (gamma 4.12, 4.13)

garethbrinn:
Draft produced by the Field Content Pipeline from **[What a gateway does that a token can't](https://choward18.github.io/gravitee-authz-reference-architectures/github-mcp-proxy.html)** by **Cole Howard, Field CTO**.

---

# Part 1 — Questions for the author

@choward18 — the pipeline verified every claim against source and couldn't settle the following from code alone. Reply in a comment, numbered to match. Prose is fine — no form.

**1. The policy language is GAPL, not Cedar.** The article says *"Policy is written in [Cedar]"* and links to `cedarpolicy.com`. The product's language is **GAPL** (Gravitee Authorization Policy Language), a *subset* of Cedar — `gravitee-gamma-module-authz` engine package `io.gravitee.authz.engine`, and the published `authorization-management-overview.md` states the subset relationship. GAPL drops some Cedar features, including `unless` clauses. Your examples are all valid GAPL, so nothing you wrote breaks — but the naming and the outbound link both need correcting in the asset. Was Cedar a deliberate simplification for a field audience, or did the GAPL naming not reach you?

**2. Was per-identity rate limiting actually configured in your build, or inferred?** The article presents identity keying as inherent to the control. In `gravitee-policy-ratelimit/schema-form.json`, the Rate Limit policy defaults to counting against the **plan and subscription pair**, not the caller. Identity keying requires setting `key` to an EL expression *and* enabling `useKeyOnly`. If your demo showed per-identity throttling, you must have set those — can you share the values? If it wasn't configured, the `bob → calls 11-13: throttled` result may have come from the plan/subscription default rather than from identity.

**3. The decision-log sample doesn't match the emitted format.** Your sample shows `action=github-mcp-server.list_issues` and `reason="Permitted by policy 'engineering-can-use-tools'"`. Source (`gravitee-policy-authz-pep`, decision-logging path) emits the **raw** tool name as the action — only the *resource* is server-qualified — and it carries the policy identifier in a separate `matchedPolicies` field, with `reason` holding the policy's own annotations. So a real line reads closer to `action=list_issues ... matchedPolicies=[engineering-can-use-tools]`. Was that sample copied from a real log or reconstructed for the article?

**4. "The engine runs inside the gateway" — true, but topology-dependent.** This one you got right, and it verified better than expected: the PEP's own config schema describes calling the **in-gateway PDP**, and it reaches it over an internal event-bus address rather than an HTTP hop. Worth flagging that a **sidecar PDP** is also a documented deployment option, so the guarantee holds for the configuration you tested rather than universally. Which were you running?

**5. Observed internal — the decision log format (rule 30 call).** The literal log line is real but isn't published contract. The draft describes *what a decision records* — subject, action, resource, decision, determining policies — without the format string, so a reader can rely on the capability without building a parser against an internal shape. **Recommendation: withhold the literal format.** Do you or @gravitee-io/docs-team disagree?

**6. Licence tier is unresolved.** The PEP declares `feature=gamma-authz-policy-pep` and the authorization module declares `feature=gamma-authz-module`. Rule 9's mapping file (`readme/enterprise-edition.md`) doesn't exist in `gravitee-platform-docs`, and the `gamma-*` prefix isn't in rule 9's documented prefix list. **The draft therefore states no licence tier at all.** Which pack are these in?

**7. Third-party cuts under rule 7 — confirm the line.** Cut and not replaced: the GitHub MCP tool inventory, PAT scope behavior, server-wide read-only mode, and all of Part 7's comparison against the platform's own controls. The draft is written server-agnostically, which matches your own closing point that "the box is replaceable." Can a Gravitee reader complete the Gravitee-side task without any of it? If you think one piece is load-bearing, say which and it goes back in as minimal integration context.

**8. A capability your framing implies doesn't exist.** Part 5 frames volume control as a single cap. Two sibling policies in the same repo also support MCP proxies and neither appears in the article: **Quota** for longer-period allowances, and **Spike Arrest** for burst smoothing. Conversely **Token Bucket Rate Limit** does *not* declare `mcp_proxy`. Worth adding to your next revision — and worth knowing before someone reaches for token-bucket on an MCP proxy.

**9. Do you have an environment I can run traffic through?** A local Docker stack with your configuration is fine. Nothing in this run was live-verified, so the following are at their source verdict only: the 403 short-circuit actually short-circuiting, per-identity throttling actually isolating callers, redaction actually removing data before egress, and decision records actually landing. Supplying one re-enters Step 5 and also produces real screenshots — right now the draft has none, and it has no placeholders either, because it's a concept page where no screenshot trigger applies.

**10. Which build did you test?** The asset states no version, product, or date. The draft targets **gamma 4.12 and 4.13** on the assumption it's current. If you tested an older build, several claims need re-verifying against these folders rather than being promoted from a patch release.

---

# Part 2 — What the pipeline did

## Source asset

| | |
|---|---|
| Asset | https://choward18.github.io/gravitee-authz-reference-architectures/github-mcp-proxy.html |
| Author | Cole Howard, Field CTO |
| Date | Not stated |
| Product / version | Not stated — inferred as Gamma, targeted at 4.12 and 4.13 |
| Self-declared verification scope | **None.** No per-claim testing status. Under Step 0 every claim is recorded as *unknown*, never as tested. |

This asset is unusual for this pipeline: a persuasive essay rather than the deep integration guide the pipeline is tuned for. Roughly half the body is bucket F framing.

## Claim verdicts

**Confirmed and shipped (11)**

| Claim | Source |
|---|---|
| MCP Studio is a mode of the MCP proxy, not a separate product | `McpProxyMode.java`, `ApimMcpStudioWriter.java` |
| Composition/curation of tools from several upstream servers | `ApimMcpStudioWriter.java`, `parseStudioTools.ts` |
| Enabling FGA wires the authorization PEP into the studio flow | `UpdateMcpStudioFgaUseCase.java`, `ApimMcpStudioWriter.FGA_PEP_CONFIG` |
| Subject is `User::"<sub>"`, resource is `MCPTool::"…"` | `FGA_PEP_CONFIG` |
| Default-deny: no matching permit means denied | authz-pep schema — denies on `FORBID` **and** `NOT_APPLICABLE` |
| `forbid` beats `permit` | engine `Authorizer.isAuthorized`, `mcp-policy-examples.md` |
| Denied call returns **403** | `FGA_PEP_CONFIG` `denyStatus: 403` |
| **No upstream call on denial** | request-phase `ctx.interruptWith(...)` in the PEP |
| Decision logging records subject, action, resource, decision, policies | authz-pep decision-logging path, `logDecisions: true` |
| **PII filtering screens responses, on by default** | `PiiFilteringPolicy.java:79-80` `ctx.addActionOnResponse(...)`, `skipResponsePayloadFiltering` default `false` |
| Redaction renders as `[REDACTED]` | `PiiConstants.java:5` |

The PII result is worth calling out. The manifest declares `mcp_proxy=REQUEST`, so reading the manifest alone would have produced a confident, wrong *"this policy can't touch responses."* Reading the policy class settled it — the same failure mode the README documents from a previous field asset, in the opposite direction.

**Corrected (4)** — questions 1, 2, 3, and the category list. The article names 6 PII categories; there are **13** (`PiiCategory.java`), and `categories` is a required setting with no default. The draft lists all 13.

**Excluded (3)** — the licence tier (question 6), the literal decision-log format (question 5, rule 30), and the "microseconds" figure. That last one is a bucket E measurement from the author's build. Note the existing `add-policies-to-mcp-server.md` already claims microsecond latency, so the number isn't new — this draft simply doesn't restate a performance figure it didn't measure.

**Absence claims, stated at the layer verified.** The only absence claim in the draft is *"Token Bucket Rate Limit doesn't support MCP proxies"*, verified at the plugin-manifest layer: `gravitee-policy-token-bucket-ratelimit/plugin.properties` declares `http_proxy` and `http_message` and no `mcp_proxy`, while its three siblings in the same repo do declare it.

## Third-party content cut (rule 7)

GitHub MCP tool inventory, PAT scope semantics, server-wide read-only mode, all of Part 7's platform comparison, and the `cedarpolicy.com` link. See question 7.

## Marketing and framing cut (bucket F)

Parts 1 and 8 in full, the "unusually enthusiastic caller" framing, "The box is replaceable", and the closing exhortations. The three-tier Admin/Engineering/Triage narrative survives only as the abstracted use cases.

## Overlap map — read this first

**Overlap is heavy. Closing this PR as already-covered is a legitimate outcome, and placement and consolidation are the writer's call, not the pipeline's.**

| Draft section | Existing coverage | How complete |
|---|---|---|
| Authorization decides which tools a caller reaches | `configure-your-mcp/add-policies-to-mcp-server.md` | **Near-total.** Already covers per-tool GAPL, principals, permit/forbid, "rejected before reaching the upstream", and microsecond/no-network-hop latency. |
| GAPL policy examples | `authorization-management/configure/mcp-policy-examples.md` | **Near-total.** Already has per-tool permit/forbid and a forbid-overrides-permit tip. |
| Rate limits decide how often | `configure-your-mcp/apply-policies-to-tool-invocations.md` | **Partial.** Documents scoping a rate limit to one tool by `mcp_tool_name`, but not which rate-limit policies support MCP proxies, and not identity keying. |
| Redaction decides what comes back | `apim/4.13/…/policy-reference/pii-filtering.md` | **Partial.** APIM policy reference exists. Nothing covers applying it to an MCP tool **response** path. |
| Curation as a control | `agent-management/build/create-an-mcp-studio.md` | **Total.** Already names PII redaction, rate limiting, and fine-grained authorization in one sentence. |

**What is genuinely new here:** only the composition — the three controls described as one layered model on a single Composite MCP Server, plus the identity-keying and MCP-support details for the rate-limit family. If the writer judges that isn't worth its own page, the rate-limit and PII specifics are the parts worth grafting onto the two partial pages.

## Images

**None, and no placeholders.** This is a concept page with no procedures, so none of the five triggers in `images-and-figures.md` applies. Step 5 had no environment regardless — see question 9.

## Docs bug found on the way

Not fixed in this PR, and worth its own issue. `add-policies-to-mcp-server.md` tells you to pick actions from `invoke`, `list`, or `read`, and `mcp-policy-examples.md` uses `action == Action::"invoke"`. The MCP visual editor emits one action per tool, mirroring the tool identifier (`Action::"<toolUid>"` in `mcp-policy-gapl.ts`), and the runtime PEP sends the raw tool name as the action. Either the docs describe an action vocabulary the editor doesn't emit, or there are two distinct layers that the pages don't distinguish. The draft sidesteps this by using `Action::"invoke"` to match the existing published examples — **if that's wrong, this draft inherits the error and both existing pages need fixing too.**

## Setup note

Two repos were missing at the start and blocked verification under rule 8. Both are now cloned and were decisive: `gravitee-policy-authz-pep` settled the 403 short-circuit, default-deny, and log format, and `gravitee-policy-ratelimit` settled the identity-keying correction and the sibling-policy findings. Worth adding both to the pipeline README's clone list.

---


garethbrinn:
Follow-up: the action-clause discrepancy flagged under **Docs bug found on the way** has now been verified and fixed in #2103.

Outcome: both published claims were wrong. The MCP visual editor has no actions picker, and `invoke`/`list`/`read` come from a read/write *tool-name* classification heuristic in `McpToolTree.tsx`, not from an action vocabulary. `Action::"invoke"` appears nowhere in product source.

This resolves the open question in this PR's description: the draft here inherited the error by using `Action::"invoke"` to match the published examples. #2103 corrects the examples to the unconstrained `action` form, so **this page should be updated to match once #2103 merges** — or rebased on it.

garethbrinn:
Correction to my earlier comment: #2103 has been scaled back.

The claim that the MCP editor has no actions picker came from reading `gravitee-gamma-module-authz` at HEAD (**1.12.0**). The 4.12.11 stack actually ships **1.2.0**, where the MCP visual editor doesn't exist and the generic policy builder *does* have an actions chip picker. The step is real, and #2103 no longer removes it.

What survives is narrower: `invoke`/`list`/`read` is not an action vocabulary at any version — actions come from the **Actions** page.

Consequence for this PR: I've **reverted the `action` clause change** on the drafted page, so it matches the published examples again. Whether `Action::"invoke"` is right for MCP still needs a licensed instance to settle — added to the author questions.

choward18:
> Draft produced by the Field Content Pipeline from **[What a gateway does that a token can't](https://choward18.github.io/gravitee-authz-reference-architectures/github-mcp-proxy.html)** by **Cole Howard, Field CTO**.
> 
> # Part 1 — Questions for the author
> @choward18 — the pipeline verified every claim against source and couldn't settle the following from code alone. Reply in a comment, numbered to match. Prose is fine — no form.
> 
> **1. The policy language is GAPL, not Cedar.** The article says _"Policy is written in [Cedar]"_ and links to `cedarpolicy.com`. The product's language is **GAPL** (Gravitee Authorization Policy Language), a _subset_ of Cedar — `gravitee-gamma-module-authz` engine package `io.gravitee.authz.engine`, and the published `authorization-management-overview.md` states the subset relationship. GAPL drops some Cedar features, including `unless` clauses. Your examples are all valid GAPL, so nothing you wrote breaks — but the naming and the outbound link both need correcting in the asset. Was Cedar a deliberate simplification for a field audience, or did the GAPL naming not reach you?
> 
> **2. Was per-identity rate limiting actually configured in your build, or inferred?** The article presents identity keying as inherent to the control. In `gravitee-policy-ratelimit/schema-form.json`, the Rate Limit policy defaults to counting against the **plan and subscription pair**, not the caller. Identity keying requires setting `key` to an EL expression _and_ enabling `useKeyOnly`. If your demo showed per-identity throttling, you must have set those — can you share the values? If it wasn't configured, the `bob → calls 11-13: throttled` result may have come from the plan/subscription default rather than from identity.
> 
> **3. The decision-log sample doesn't match the emitted format.** Your sample shows `action=github-mcp-server.list_issues` and `reason="Permitted by policy 'engineering-can-use-tools'"`. Source (`gravitee-policy-authz-pep`, decision-logging path) emits the **raw** tool name as the action — only the _resource_ is server-qualified — and it carries the policy identifier in a separate `matchedPolicies` field, with `reason` holding the policy's own annotations. So a real line reads closer to `action=list_issues ... matchedPolicies=[engineering-can-use-tools]`. Was that sample copied from a real log or reconstructed for the article?
> 
> **4. "The engine runs inside the gateway" — true, but topology-dependent.** This one you got right, and it verified better than expected: the PEP's own config schema describes calling the **in-gateway PDP**, and it reaches it over an internal event-bus address rather than an HTTP hop. Worth flagging that a **sidecar PDP** is also a documented deployment option, so the guarantee holds for the configuration you tested rather than universally. Which were you running?
> 
> **5. Observed internal — the decision log format (rule 30 call).** The literal log line is real but isn't published contract. The draft describes _what a decision records_ — subject, action, resource, decision, determining policies — without the format string, so a reader can rely on the capability without building a parser against an internal shape. **Recommendation: withhold the literal format.** Do you or @gravitee-io/docs-team disagree?
> 
> **6. Licence tier is unresolved.** The PEP declares `feature=gamma-authz-policy-pep` and the authorization module declares `feature=gamma-authz-module`. Rule 9's mapping file (`readme/enterprise-edition.md`) doesn't exist in `gravitee-platform-docs`, and the `gamma-*` prefix isn't in rule 9's documented prefix list. **The draft therefore states no licence tier at all.** Which pack are these in?
> 
> **7. Third-party cuts under rule 7 — confirm the line.** Cut and not replaced: the GitHub MCP tool inventory, PAT scope behavior, server-wide read-only mode, and all of Part 7's comparison against the platform's own controls. The draft is written server-agnostically, which matches your own closing point that "the box is replaceable." Can a Gravitee reader complete the Gravitee-side task without any of it? If you think one piece is load-bearing, say which and it goes back in as minimal integration context.
> 
> **8. A capability your framing implies doesn't exist.** Part 5 frames volume control as a single cap. Two sibling policies in the same repo also support MCP proxies and neither appears in the article: **Quota** for longer-period allowances, and **Spike Arrest** for burst smoothing. Conversely **Token Bucket Rate Limit** does _not_ declare `mcp_proxy`. Worth adding to your next revision — and worth knowing before someone reaches for token-bucket on an MCP proxy.
> 
> **9. Do you have an environment I can run traffic through?** A local Docker stack with your configuration is fine. Nothing in this run was live-verified, so the following are at their source verdict only: the 403 short-circuit actually short-circuiting, per-identity throttling actually isolating callers, redaction actually removing data before egress, and decision records actually landing. Supplying one re-enters Step 5 and also produces real screenshots — right now the draft has none, and it has no placeholders either, because it's a concept page where no screenshot trigger applies.
> 
> **10. Which build did you test?** The asset states no version, product, or date. The draft targets **gamma 4.12 and 4.13** on the assumption it's current. If you tested an older build, several claims need re-verifying against these folders rather than being promoted from a patch release.
> 
> # Part 2 — What the pipeline did
> ## Source asset
> Asset	https://choward18.github.io/gravitee-authz-reference-architectures/github-mcp-proxy.html
> Author	Cole Howard, Field CTO
> Date	Not stated
> Product / version	Not stated — inferred as Gamma, targeted at 4.12 and 4.13
> Self-declared verification scope	**None.** No per-claim testing status. Under Step 0 every claim is recorded as _unknown_, never as tested.
> This asset is unusual for this pipeline: a persuasive essay rather than the deep integration guide the pipeline is tuned for. Roughly half the body is bucket F framing.
> 
> ## Claim verdicts
> **Confirmed and shipped (11)**
> 
> Claim	Source
> MCP Studio is a mode of the MCP proxy, not a separate product	`McpProxyMode.java`, `ApimMcpStudioWriter.java`
> Composition/curation of tools from several upstream servers	`ApimMcpStudioWriter.java`, `parseStudioTools.ts`
> Enabling FGA wires the authorization PEP into the studio flow	`UpdateMcpStudioFgaUseCase.java`, `ApimMcpStudioWriter.FGA_PEP_CONFIG`
> Subject is `User::"<sub>"`, resource is `MCPTool::"…"`	`FGA_PEP_CONFIG`
> Default-deny: no matching permit means denied	authz-pep schema — denies on `FORBID` **and** `NOT_APPLICABLE`
> `forbid` beats `permit`	engine `Authorizer.isAuthorized`, `mcp-policy-examples.md`
> Denied call returns **403**	`FGA_PEP_CONFIG` `denyStatus: 403`
> **No upstream call on denial**	request-phase `ctx.interruptWith(...)` in the PEP
> Decision logging records subject, action, resource, decision, policies	authz-pep decision-logging path, `logDecisions: true`
> **PII filtering screens responses, on by default**	`PiiFilteringPolicy.java:79-80` `ctx.addActionOnResponse(...)`, `skipResponsePayloadFiltering` default `false`
> Redaction renders as `[REDACTED]`	`PiiConstants.java:5`
> The PII result is worth calling out. The manifest declares `mcp_proxy=REQUEST`, so reading the manifest alone would have produced a confident, wrong _"this policy can't touch responses."_ Reading the policy class settled it — the same failure mode the README documents from a previous field asset, in the opposite direction.
> 
> **Corrected (4)** — questions 1, 2, 3, and the category list. The article names 6 PII categories; there are **13** (`PiiCategory.java`), and `categories` is a required setting with no default. The draft lists all 13.
> 
> **Excluded (3)** — the licence tier (question 6), the literal decision-log format (question 5, rule 30), and the "microseconds" figure. That last one is a bucket E measurement from the author's build. Note the existing `add-policies-to-mcp-server.md` already claims microsecond latency, so the number isn't new — this draft simply doesn't restate a performance figure it didn't measure.
> 
> **Absence claims, stated at the layer verified.** The only absence claim in the draft is _"Token Bucket Rate Limit doesn't support MCP proxies"_, verified at the plugin-manifest layer: `gravitee-policy-token-bucket-ratelimit/plugin.properties` declares `http_proxy` and `http_message` and no `mcp_proxy`, while its three siblings in the same repo do declare it.
> 
> ## Third-party content cut (rule 7)
> GitHub MCP tool inventory, PAT scope semantics, server-wide read-only mode, all of Part 7's platform comparison, and the `cedarpolicy.com` link. See question 7.
> 
> ## Marketing and framing cut (bucket F)
> Parts 1 and 8 in full, the "unusually enthusiastic caller" framing, "The box is replaceable", and the closing exhortations. The three-tier Admin/Engineering/Triage narrative survives only as the abstracted use cases.
> 
> ## Overlap map — read this first
> **Overlap is heavy. Closing this PR as already-covered is a legitimate outcome, and placement and consolidation are the writer's call, not the pipeline's.**
> 
> Draft section	Existing coverage	How complete
> Authorization decides which tools a caller reaches	`configure-your-mcp/add-policies-to-mcp-server.md`	**Near-total.** Already covers per-tool GAPL, principals, permit/forbid, "rejected before reaching the upstream", and microsecond/no-network-hop latency.
> GAPL policy examples	`authorization-management/configure/mcp-policy-examples.md`	**Near-total.** Already has per-tool permit/forbid and a forbid-overrides-permit tip.
> Rate limits decide how often	`configure-your-mcp/apply-policies-to-tool-invocations.md`	**Partial.** Documents scoping a rate limit to one tool by `mcp_tool_name`, but not which rate-limit policies support MCP proxies, and not identity keying.
> Redaction decides what comes back	`apim/4.13/…/policy-reference/pii-filtering.md`	**Partial.** APIM policy reference exists. Nothing covers applying it to an MCP tool **response** path.
> Curation as a control	`agent-management/build/create-an-mcp-studio.md`	**Total.** Already names PII redaction, rate limiting, and fine-grained authorization in one sentence.
> **What is genuinely new here:** only the composition — the three controls described as one layered model on a single Composite MCP Server, plus the identity-keying and MCP-support details for the rate-limit family. If the writer judges that isn't worth its own page, the rate-limit and PII specifics are the parts worth grafting onto the two partial pages.
> 
> ## Images
> **None, and no placeholders.** This is a concept page with no procedures, so none of the five triggers in `images-and-figures.md` applies. Step 5 had no environment regardless — see question 9.
> 
> ## Docs bug found on the way
> Not fixed in this PR, and worth its own issue. `add-policies-to-mcp-server.md` tells you to pick actions from `invoke`, `list`, or `read`, and `mcp-policy-examples.md` uses `action == Action::"invoke"`. The MCP visual editor emits one action per tool, mirroring the tool identifier (`Action::"<toolUid>"` in `mcp-policy-gapl.ts`), and the runtime PEP sends the raw tool name as the action. Either the docs describe an action vocabulary the editor doesn't emit, or there are two distinct layers that the pages don't distinguish. The draft sidesteps this by using `Action::"invoke"` to match the existing published examples — **if that's wrong, this draft inherits the error and both existing pages need fixing too.**
> 
> ## Setup note
> Two repos were missing at the start and blocked verification under rule 8. Both are now cloned and were decisive: `gravitee-policy-authz-pep` settled the 403 short-circuit, default-deny, and log format, and `gravitee-policy-ratelimit` settled the identity-keying correction and the sibling-policy findings. Worth adding both to the pipeline README's clone list.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
Re-verified all of this live today against the running stack before answering — versions in (10). Where you're right I've said so plainly; two items came back different from your source reading, and I've included the raw evidence for those.

1. GAPL, not Cedar — you're right. The docs say it explicitly (docs/gamma/4.12/authorization-management/get-started/authorization-management-overview.md): "Policies are written in GAPL (Gravitee Authorization Policy Language), a subset of the Cedar policy language", and "Cedar features like unless clauses are not available in the GAPL editor." The product says it too — the plugin's own name is Gravitee Authorization PEP (GAPL), class GaplAuthzPolicy. And the naming did reach me: my own demo runbook says "GAPL when conditions". I wrote Cedar in the asset for reader recognizability, which is defensible for the syntax and indefensible for the link — cedarpolicy.com documents features GAPL doesn't have, so it actively misleads. Fix it to "GAPL (a subset of Cedar)" and point the link at the Gravitee authorization-management docs. **GAPL naming. Applied. "Policy is written in [Cedar]" with the cedarpolicy.com link became "Policy is written in GAPL — the Gravitee Authorization Policy Language, a subset of Cedar", and the outbound link is gone entirely rather than repointed. The code comment "In Cedar, forbid always beats permit" is now "In GAPL". Verified zero residual cedarpolicy or "written in Cedar" strings.**

2. Configured, not inferred — here are the values. From the live deployed definition on studio abaae37e:

{"addHeaders": true, "async": false,
 "rate": {"limit": 10, "periodTime": 60, "periodTimeUnit": "SECONDS",
  "key": "{#context.attributes['user']}", "useKeyOnly": false},
 "errorStrategy": "FALLBACK_PASS_TROUGH"}

key is set to the identity EL, so this is not the plan/subscription default. But your instinct to check was right, because useKeyOnly: false means the counter is composite — plan and subscription and the key — not purely per-identity.

I ran the decisive experiment rather than argue it. Both personas authenticate through the same policy-verifier client, so if the counter were subscription-scoped, the second caller would inherit the first's exhausted budget:

bob   calls  1–10 : allowed        (remaining 9 → 0)
bob   calls 11–13 : THROTTLED 429
trish calls  1–3  : allowed        (remaining 9, 8, 7)   ← fresh counter

Isolation holds. So the article's result is real. The precise wording, though, should be "per identity within a subscription" — useKeyOnly: true is what makes it identity-scoped regardless of plan and subscription, and I'd argue that's what the demo should actually set. Flagging that as a change I want to make, not just a caveat to write around.

**Rate-limit framing. Softened, not removed, since the behavior verified. Added a qualifier under the rate-limit intro: the counter follows the subscription by default, and pointing it at the caller is a deliberate configuration choice you don't inherit. That makes the claim true without publishing useKeyOnly internals.**


3. The sample is real and abridged, not reconstructed — but you're half right. Verbatim from the gateway just now:

authz decision requestId=… env=DEFAULT org=DEFAULT apiId=abaae37e-…
  subject=User::"90fe50a7-…" action=github-mcp-server.create_or_update_file
  resource=MCPTool::"github-mcp-server.create_or_update_file" decision=FORBID
  matchedPolicies=[0a780976-…#0,10538fad-…#0]
  reason="Forbidden by policy 'engineering-forbid-create_or_update_file'" durationNs=208208

You're right that matchedPolicies exists as a separate field — I omitted it, along with requestId, env, org, apiId and durationNs. The original asset did say identifying UUIDs and the timing field were trimmed.

Two corrections. action is server-qualified here, not the raw tool name — because actionExpr is a configurable EL and this deployment sets it to {#context.attributes['mcp_tool_name']}, the same expression as resourceExpr. If your source reading was of the default, that's the reconciliation; the field is configuration-dependent, so neither of us should state it as fixed behavior. And reason does carry the policy name in this build — "Permitted by policy 'engineering-can-use-tools'" is the literal emitted string, which is what I published.

One bonus that supports the article's "forbid beats permit" claim: on a FORBID, matchedPolicies lists both the permit and the forbid.

**3 + 5 — Decision log. Taken together, since 5's recommendation resolves 3. The literal log block is deleted. In its place, prose naming what a decision records — calling identity, tool, decision, determining policy, evaluation duration — plus a new paragraph on why the determining-policy field is the one people underestimate. Nobody can build a parser against what's published now.

I also added a short callout making the FORBID vs default-deny distinction reader-facing, since your earlier question showed it wasn't obvious, and today's run gave me ground truth for it (NOT_APPLICABLE, matchedPolicies=[], "No applicable policy").**


4. In-gateway, and I ran exactly that configuration. The plugin text is unambiguous. PEP schema: "Per-request guard that calls the in-gateway PDP (compiled GAPL bundle)." PDP service: "hosts one or more GAPL Policy Decision Point engines in-process… pushed over the Vert.x EventBus… answers eval and search requests on the service:authz-pdp addresses." gravitee-service-authz-pdp-1.0.1 is installed in the gateway container and my PEP config carries no remote endpoint. Agree the claim should be scoped to the tested configuration. For completeness, the alternative path in this build is the AuthZEN REST route (gravitee-policy-authz-pdp-authzen, gravitee-entrypoint-authz) — I have not tested it, and my runbook notes the authzen entrypoint doesn't bind on this image.

5. No disagreement — withhold the format. It's observed internal behavior and I'd rather nobody build a parser against it. Describing what a decision records — subject, action, resource, decision, determining policy, duration — carries the whole argument without pinning the shape. Your call and the docs team's; I'm not attached to the literal line.
**see if for 3**

6. My licence reports tier: universe with packs including authorization-management and agent-management. The name correspondence is suggestive but it's inference from strings, not documentation — so keep the draft stating no tier, and let product/licensing supply the authoritative mapping.

7. Agree with every cut; nothing removed is load-bearing. A Gravitee reader can complete the Gravitee-side task without the tool inventory, PAT scopes, or read-only mode — those were rhetorical contrast, not instructions. One thing I'd ask back, and it can be stated vendor-neutrally: the premise that a shared upstream credential grants all its holders the same authority. Without it the reader has no reason the gateway layer exists. That's a general property of bearer credentials, so it should survive rule 7 without naming anyone.

8. Fair catch, and confirmed from the manifests in the running gateway. rate-limit 5.0.0, quota 5.0.0 and spike-arrest 5.0.0 all declare mcp_proxy=REQUEST. token-bucket-rate-limit 5.0.0 declares only http_proxy and http_message — no mcp_proxy. So quota and spike arrest go into the revision, and the token-bucket trap is worth calling out explicitly, because it's the one someone reaches for when they want burst tolerance.

9. Yes — local Docker stack, and it's up now. Studio engineering-toolbelt on localhost:8082, three personas, AM tokens from localhost:8092. All four of your source-verdict items are live-verified as of today: the 403 short-circuit (create_or_update_file as bob → 403 FORBID; get_file_contents as trish → 403 NOT_APPLICABLE, matchedPolicies=[], "No applicable policy" — worth noting those are two distinct deny paths); throttle isolation as shown in (2); redaction (a real name, email and phone come back [REDACTED] with the non-PII line intact); and decision records landing. Happy to hand over the drivers or run whatever sequence you want for screenshots.

One thing to know before you shoot a successful-write screenshot: the studio's GitHub upstream auth currently returns "Authorization header badly formatted", so alice's permitted write doesn't execute end-to-end. The deny path is unaffected — bob is refused before the upstream — but the happy path needs the credential fixed first.

10. Build. APIM gateway and management API 4.12.12-SNAPSHOT (images 4.12.x-latest), gamma-module-authz 1.2.0, gamma-module-aim 1.5.2, policy-authz-pep 1.1.0, service-authz-pdp 1.0.1, policy-ratelimit 5.0.0, policy-pii-filtering 2.2.0, MCP entrypoint/endpoint 3.1.1, AM master-latest. So targeting 4.12 is correct for everything above — I have not verified anything on 4.13. Worth noting aim is a released 1.5.2 here; it was an alpha earlier in July, so any older note describing it that way is stale.


Doc is live with stated changes

choward18:
## Step 10 — answers incorporated, and Step 5 re-entered

Question 9 supplied the environment, so per Step 10 I re-entered Step 5 rather than just recording the answers. Everything below was verified live on a licensed instance this run — **APIM 4.12.12-SNAPSHOT, `gamma-module-authz` 1.2.0, `policy-authz-pep` 1.1.0, `service-authz-pdp` 1.0.1, licence `tier: universe`** — not inferred from my answers.

Per the step's own rule, I treated my own answers as leads and not evidence. That was the right call: **one of my answers was incomplete and one of the pipeline's source readings was wrong**, and only traffic settled either.

---

### 🔴 Blocking — every clause of the drafted policy example fails to match

This is the most consequential finding of the run and it supersedes the `Action::"invoke"` discussion in the two comments above. I tested each clause as an **isolated arm** — create policy → deploy → wait for sync → call → delete → confirm the deny returns — against a principal (`trish`/Triage) who is otherwise default-denied the target tool. A control arm proves the harness flips the decision.

| Clause as drafted | Result | Control |
|---|---|---|
| `principal in Group::"engineering"` (display name) | **NOT_APPLICABLE** | `Group::"<group-id>"` → PERMIT |
| `action == Action::"invoke"` | **NOT_APPLICABLE** | unconstrained `action` → PERMIT |
| `resource in MCPServer::"github-mcp-server"` | **NOT_APPLICABLE** | `resource == MCPTool::"<server>.<tool>"` → PERMIT |

My first attempt at this test returned a false positive because the arms were cumulative — the control policy was still deployed when the later arms ran, so all three reported PERMIT. Re-running with each arm created and deleted in isolation produced the table above. Flagging the methodology because the wrong version of this test would have confirmed the draft.

**All three clauses are wrong, so the drafted `permit` matches nothing.** A reader who copies it gets default-deny and their tools stop working, with no error explaining why — the failure is silent by design, which is exactly what makes it worth blocking on.

This settles the question left open in the second comment above: `Action::"invoke"` is not correct for MCP at 4.12 with `gamma-module-authz` 1.2.0. The narrower conclusion there — that `invoke`/`list`/`read` is not an action vocabulary — holds and now has runtime evidence behind it.

**Downstream impact.** The draft was deliberately matched to the published examples, so `add-policies-to-mcp-server.md` and `mcp-policy-examples.md` carry the same non-matching pattern. That's a docs bug affecting shipped pages, not just this PR, and it's larger than the one already scoped into #2103.

---

### Corrections to the drafted page

**Question 2 — the `useKeyOnly` guidance is contradicted by traffic.** The page currently says that without **Use key only**, *"several identities on one subscription share a single allowance."* That is not what happens.

My deployed config sets `key` to the caller EL with `useKeyOnly: false`. Two identities on the **same** `policy-verifier` client:

```
bob   calls  1–10 : allowed        (remaining 9 → 0)
bob   calls 11–13 : 429
trish calls  1–3  : allowed        (remaining 9, 8, 7)   ← independent counter
```

Setting **Key** alone is what produces per-identity budgets. **Use key only** additionally drops the plan and subscription from the counter, which matters across subscriptions rather than within one. Corrected text below.

This also corrects my own answer 2, where I described the isolation as holding "within a subscription" — accurate, but I attributed more to `useKeyOnly` than it does.

**Questions 1, 3, 4, 5, 7 — no change to the page.** All five were already handled correctly:

1. **GAPL** — the page already names it correctly and carries no `cedarpolicy.com` link. Confirmed against the published `authorization-management-overview.md` and the plugin's own name, `Gravitee Authorization PEP (GAPL)`. I've corrected the source asset.
3 + 5. **Decision log** — correctly excluded under rule 30. For the record, the emitted line does carry `matchedPolicies` as a separate field, and `action` is server-qualified in my build only because `actionExpr` is a configurable EL set to the tool attribute. Neither belongs on a page.
4. **Topology** — the page's in-gateway framing is right. Verified: the PEP schema describes calling the in-gateway PDP, and the PDP service hosts engines in-process over the Vert.x EventBus on `service:authz-pdp`. The AuthZEN REST path exists in this build but I did **not** test it, so nothing should claim universality.
7. **Third-party cuts** — all confirmed. A Gravitee reader completes the task without any of it. The one premise worth keeping is that a shared upstream credential grants all its holders the same authority, which is a property of bearer credentials generally and needs no vendor named. The page's opening sentence already covers it.

**Question 6 — licence still excluded, but here are the strings.** The mapping genuinely isn't published: `readme/enterprise-edition.md` 404s and `docs/ee/license.md` is install instructions only. Declared features: `gamma-authz-module` (1.2.0), `gamma-authz-policy-pep` (1.1.0), `gamma-authz-service-pdp` (1.0.1), `gamma-authz-reactor`, `gamma-aim-module` (1.5.2). Note **PII Filtering is not in that family** — it declares `apim-policy-pii-filtering`, so the page's three controls do not sit in one pack. My licence reports packs including `authorization-management` and `agent-management`. The name correspondence is suggestive and is not documentation, so I'd keep the tier off the page.

**Question 8 — confirmed from the manifests in the running gateway.** `rate-limit`, `quota` and `spike-arrest` 5.0.0 all declare `mcp_proxy=REQUEST`; `token-bucket-rate-limit` 5.0.0 declares only `http_proxy` and `http_message`. The page's absence claim is correct at the manifest layer.

**Question 10 — version.** Everything above is 4.12 on the build named at the top. **Nothing here is verified on 4.13**, and the two version folders carry identical copies. Worth deciding whether 4.13 ships unverified or gets its own pass.

---

### Verified unchanged

These were live-confirmed this run and need no edit: default-deny (`NOT_APPLICABLE`, `matchedPolicies=[]`), `forbid` beating `permit` (the FORBID record lists both the permit and the forbid), `403` on denial with no upstream call, decision logging recording subject/action/resource/decision/policies, PII filtering screening responses by default and rendering `[REDACTED]` while non-PII content passes through, and the 13-category enum being required with no default — the page's list matches the schema exactly.

---

### Proposed diff

I don't have write access to `gravitee-io/gravitee-platform-docs`, so I can't push to `doc-update/mcp-layered-governance` myself. The corrected text is below for whoever holds the branch; it applies identically to the 4.12 and 4.13 copies.

**Replace the policy example and the warning hint:**

```
permit (
  principal in Group::"<group-id>",
  action,
  resource
);

forbid (
  principal in Group::"<group-id>",
  action,
  resource == MCPTool::"github-mcp-server.create_or_update_file"
);
```

Each entity reference in a policy is a unique identifier, not a display name:

* **`principal`** takes the group's identifier, not the group's name.
* **`resource`** takes the server-qualified tool identifier, in the form `MCPTool::"<server>.<tool>"`.
* **`action`** takes the same server-qualified tool identifier that the gateway sends at runtime. Leave `action` unconstrained unless you intend to match one specific tool invocation.

A clause that references a name where an identifier is expected doesn't match. Because an unmatched call is denied rather than permitted, the mistake surfaces as an unexplained denial.

> **Hint (warning):** Constrain a policy only on values the gateway actually sends. A clause that constrains `action` or `resource` to any other vocabulary matches nothing, and the call is then denied by default rather than reported as a malformed policy.

**Replace the `useKeyOnly` paragraph:**

> By default, the Rate Limit policy counts against the plan and subscription pair rather than the caller. To give each identity its own budget, set **Key** to an expression that resolves the caller's identity. Each identity then gets its own allowance, so one runaway agent doesn't consume everyone else's headroom.
>
> **Use key only** additionally removes the plan and subscription from the counter, so a single identity draws on one allowance no matter which subscription the call arrives on. Leave it disabled to keep separate allowances per subscription.

---

### Still open

* **The entity-identifier finding needs a writer decision on scope.** It invalidates examples on two published pages, so fixing only this draft leaves the shipped ones wrong.
* **4.13 is unverified.** I have no 4.13 instance.
* **Screenshots.** Still none. The page remains concept-only with no procedure, so no trigger in `images-and-figures.md` applies — the environment being available doesn't change that.
* **Whether a readable group name resolves anywhere.** I proved the ID works and the name doesn't at the PEP; I did not investigate whether some other layer accepts names and translates them. Stated at the layer verified, per rule 29.

Environment stays up if anything else needs traffic run through it.

choward18:
**Branch with the corrections applied:** `choward18:doc-update/mcp-layered-governance-fixes` ([compare](https://github.com/gravitee-io/gravitee-platform-docs/compare/doc-update/mcp-layered-governance...choward18:doc-update/mcp-layered-governance-fixes)) — commit `50a482e`, branched off this PR's head, both version folders updated.

Cherry-pick it onto `doc-update/mcp-layered-governance` rather than merging it as a second PR, so this PR stays the single artifact per the pipeline's Step 10.

garethbrinn:
Cherry-picked @choward18's corrections onto this branch as 7150cc0aa (authorship preserved), rather than opening a second PR, so this stays the single artifact per the pipeline's Step 10. Clean apply on top of 9fbcfea25 — both version folders updated and still byte-identical to each other.

Thanks for running this live, Cole — the isolated per-clause tests with control arms are exactly what this page needed. Flagging three things for reviewers before this merges.

### 1. This is the third pass at the same `action` clause

The clause has flipped twice already on this branch and once on main:

- b0ab25784 — switched to the unconstrained `action` form
- 67057a3db — reverted it, "matching the published examples"
- #2103 (now on main via 0b61e8097) — backed out its own `mcp-policy-examples.md` change, reasoning: *"Whether `Action::\"invoke\"` matches at runtime depends on the deployed authz-pep/AIM versions, which this run did not reconcile."*

This commit asserts the opposite of the current state, and it's the best-evidenced pass so far — it names `gamma-module-authz` 1.2.0, which is the version #2103 concluded actually ships, and reading the wrong module version (HEAD 1.12.0) is precisely what invalidated the first attempt. But given the history, I'd like a second person to confirm before we treat it as settled.

### 2. The two MCP pages now contradict each other

This page now tells readers that an `action == Action::"invoke"` clause matches nothing and fails silently into default-deny. Meanwhile `authorization-management/configure/mcp-policy-examples.md` still ships four `action == Action::"invoke"` examples, in both 4.12 and 4.13 (lines 29, 44, 62, 77).

If the finding here is right, those examples are equally uncopyable and equally silent about it. If it's wrong, this page is the outlier. Either way we currently ship both readings of the same question, on the exact point this page warns readers about.

The other occurrences in gamma 4.12/4.13 — `ai-policy-example.md`, `create-an-action.md`, `action-naming-and-scope.md`, `create-update-delete-policies.md` — are standard-policy-editor pages, where #2103 found the actions picker is real. Those look correct as-is. `mcp-policy-examples.md` is the one in scope.

### 3. Two claims still need verification against source

Per `backlog-agent/critical-rules.md`, a commit message is a lead, not a source, so the test log isn't something the page can rest on:

- **Version coverage.** The run was on APIM 4.12.12-SNAPSHOT. This change ships to 4.12 *and* 4.13; 4.13 isn't covered by it. Snapshot rather than release is worth noting too.
- **The rate-limit edit is a separate claim.** It reverses prior guidance — previously "set **Key**, then enable **Use key only**", now "**Key** alone gives each identity its own allowance, and **Use key only** additionally drops the plan and subscription." That's an independent behavioral assertion riding in the same commit and needs its own confirmation.

Also noting for the record that this commit is co-authored by Claude, so `backlog-agent/agent-draft-review.md` applies — no self-merge, human review required.

@choward18 — could you confirm whether your 4.12 run also covers `mcp-policy-examples.md`? If the finding holds there, I'll extend the fix so the two MCP pages agree.

choward18:
Fair challenge on all three, and the version-reconciliation point in #1 is the right thing to have stopped on. I've gone back and settled it at the **source** layer rather than adding a fourth runtime observation to the pile.

## 1. Which component decides the action UID — reconciled

The reason this clause has flipped three times is that everyone (me included) has been reasoning about the wrong component. **The action UID is not decided by `authz-pep` or by `gamma-module-authz`. It's decided by whichever module writes the PEP configuration into the studio flow**, because `actionExpr` is a configurable EL and the PEP simply evaluates whatever it was given.

That module is AIM. From the shipped artifact on this stack — `gravitee-gamma-module-aim-1.5.2.jar`, class `com/graviteesource/gamma/module/aim/infra/service_provider/mcp/ApimMcpStudioWriter.class` — the constants are:

```
actionExpr    {#context.attributes['mcp_tool_name']}
resourceExpr  {#context.attributes['mcp_tool_name']}
subjectExpr   {#context.attributes['user']}
subjectType   User
resourceType  MCPTool
```

So for **any** FGA-enabled MCP Studio at AIM 1.5.2, the action UID is the MCP tool name — by construction, not by my configuration. I hadn't hand-edited the studio; that config is what the product writes when you toggle FGA on. `Action::"invoke"` cannot match, and no `authz-pep` version changes that, because the PEP isn't choosing the value.

That's the reconciliation #2103 said its run didn't have. **The version to name is AIM, not authz.** Mine is 1.5.2. If a different AIM version writes a different `actionExpr`, the answer changes — and it's now a one-line check against that constant rather than another round of runtime tests.

**This also dissolves the apparent conflict with #2103.** Its finding that the actions picker is real for the standard policy builder and mine that `Action::"invoke"` never matches for MCP are both true, because they're about different config paths. MCP Studios get a hardcoded tool-name `actionExpr`; standard APIs get whatever the builder sets. The two pages aren't contradicting each other about one mechanism — they're describing two.

## 2. Yes — the finding covers `mcp-policy-examples.md`

I tested the published Example 1 shape verbatim, isolated arm with a control, same method as before:

| Policy under test | Decision |
|---|---|
| `permit(principal in Group::"triage", action == Action::"invoke", resource in MCPServer::"github-mcp-server")` | **NOT_APPLICABLE** |
| same intent, verified forms: `permit(principal in Group::"<group-id>", action, resource)` | PERMIT |
| `permit(principal == User::"trish", …)` — Example 2's principal form | **NOT_APPLICABLE** |
| `permit(principal == User::"<sub>", …)` | PERMIT |

So all four examples carry at least one non-matching clause, and Examples 1 and 4 carry three. Please do extend the fix — you're right that shipping both readings on this exact point is the worst outcome.

**One boundary I have to hold.** Example 3 uses `AgentIdentity::"research-bot"`. I did **not** test that entity type and have no evidence about it either way. Please don't let my finding be read as covering it — it needs its own check, and by rule 29 a result about `Group` and `User` isn't evidence about `AgentIdentity`.

## 3. The two claims, stated at the layer each was actually verified

**Version coverage — you're right, and I'd state it more narrowly than my commit did.** The run is APIM `4.12.12-SNAPSHOT`, a snapshot rather than a release. **4.13 is not covered by anything I did** and I have no 4.13 instance. One nuance worth having: the decisive artifact for #1 is `gravitee-gamma-module-aim` **1.5.2**, which is a released module version even though the APIM build around it is a snapshot — so that constant isn't a snapshot artifact. Whether 4.13 ships the same AIM version is the question to answer before this text goes into the 4.13 folder, and it's answerable by reading the same constant.

**The rate-limit edit — correctly flagged as a separate claim, and it's the weaker half of my commit.** Splitting it honestly:

* *"Setting **Key** alone gives each identity its own allowance"* — **live-verified.** Two identities on the same `policy-verifier` client, `useKeyOnly: false`: one throttled at call 11 while the other immediately got a fresh counter at `remaining=9`. Repeatable.
* *"**Use key only** additionally drops the plan and subscription from the counter"* — **not verified by me.** It rests on the policy's own schema description (`"Only uses the custom key to identify the consumer, regardless of the subscription and plan"`, and `key`'s *"leave it empty to use the default behavior (plan/subscription pair)"*). I did not test `useKeyOnly: true`, and I did not test across two subscriptions — which is the only configuration where the difference is observable.

I traced `KeyFactory` in `gravitee-ratelimit-shared-5.0.0.jar` to try to settle the composition from source and could not read the exact key ordering cleanly out of the bytecode, so I'm not going to assert it. If you want that second sentence to carry the same weight as the first, it needs either a two-subscription test or someone reading `KeyFactory` in the actual repo. Happy to run the former if it's worth the setup — otherwise I'd suggest softening that sentence to describe the setting's purpose rather than asserting its mechanism.

What is safe to say either way is that the **previous** guidance was wrong: `useKeyOnly` is not a prerequisite for per-identity budgets, and readers following it would have concluded they needed it to avoid a shared allowance they were never going to have.

---

Agreed on `agent-draft-review.md` — no self-merge, and this needs a human on it regardless of how the evidence lands. Environment is still up if you want any of the above re-run or extended to `AgentIdentity`.

tolaore (approved):

PR #2107, opened 2026-07-31
docs(gamma): verify and correct the A2A Proxy exposure guide (4.12 + 4.13)

garethbrinn:
## What this is

A Doc Verification Pipeline run on `docs/gamma/4.12/agent-management/build/expose-agent-with-a2a-proxy.md`.

Verified against a live Gamma 4.12 console (Docker, EE licence) by walking the documented procedure end to end and creating a real A2A proxy through the wizard. Claims with a code-level counterpart were cross-checked against the components shipped in the 4.12 distribution.

`docs/gamma/4.13/.../expose-agent-with-a2a-proxy.md` was byte-identical before the edit, so it receives the same change in this PR.

## Verdict table

| # | Claim | Code truth | Live truth | Verdict |
|---|---|---|---|---|
| 1 | The proxy implements A2A by serving a `/.well-known/agent.json` descriptor | The gateway answers `/.well-known/agent-card.json`, and also accepts the legacy `/.well-known/agent.json` path and an extended agent card path | Not determined — no A2A upstream agent in this stack, so no card could be fetched | **Fix** — use `agent-card.json` |
| 2 | "Skill discovery" — the proxy serves the descriptor advertising agent capabilities | The gateway takes the upstream agent's card and rewrites the URLs it advertises so callers are routed back through the gateway | Not determined | **Fix** — reworded to describe pass-through and URL rewriting |
| 3 | "Per-skill authorization" — the proxy publishes each discovered skill to Authorization Management as a cataloged entity | An A2A proxy carries only a name, context path, target URL, and status. Nothing in the shipped 4.12 components relates A2A proxies to skills or to Authorization Management entities | The wizard never discovers or asks for skills; the created proxy has no skills view | **Contradicted — removed** |
| 4 | "How skill publishing works" — the 4-step publish-and-evaluate sequence | Same as row 3 | Same as row 3 | **Contradicted — removed** |
| 5 | Skill-level authorization is "structurally different from tool-level authorization on MCP Proxies" | Depends on row 3 | Depends on row 3 | **Contradicted — removed** |
| 6 | Wire-level governance through the AI Gateway with observability, authentication, and policy enforcement | Plans and policy attachment exist for A2A proxies | Confirmed — the proxy has Policy Studio, Dashboard, Logs, and Tracing, and a published plan | **Confirmed** |
| 7 | From the Gamma console sidebar, select **Agent Management** | n/a | Confirmed | **Confirmed** |
| 8 | Navigate to **Build** and select **A2A Proxies** | n/a | The Agent Management sidebar groups are General, Catalog, Secure, and Observability. **A2A Proxies** is under **Secure**. There is no Build group | **Fix** |
| 9 | Select **Create A2A proxy** to launch the wizard | n/a | Confirmed — button label matches exactly | **Confirmed** |
| 10 | **Step 1: Define** — provide a name and optional description | n/a | Step 1 is labelled **Define**. The panel asks for **Name** (required), **Context path** (required), and **Description** (optional) | **Fix** — **Context path** was undocumented |
| 11 | **Step 2: Connect** — enter the target A2A endpoint URL | n/a | Step 2 is **Secure**, not Connect | **Fix** — step order |
| 12 | **Step 3: Secure** — attach security plans | n/a | Step 3 is **Connect**, not Secure | **Fix** — step order |
| 13 | The Secure step content | The only client authentication types an A2A proxy accepts at creation are API key and mTLS | The step offers exactly **API Key** and **mTLS**, and states that a default plan is created and published with the proxy | **Fix** — options now named |
| 14 | The Connect step content | n/a | The step is "Connect to upstream agent": **Target URL** (required), plus an upstream authentication choice of **Static credential** or **No upstream auth** | **Fix** — upstream authentication was undocumented |
| 15 | "The gateway will resolve its configuration by fetching `/.well-known/agent-card.json`" during creation | The card is read and rewritten while proxying a request, not while creating the proxy | The wizard accepted an unreachable target URL and created the proxy without fetching anything | **Fix — removed** |
| 16 | **Step 4: Review** — confirm the settings and select **Create A2A proxy** | n/a | Confirmed — "Review configuration" summary, button reads **Create A2A proxy** | **Confirmed** |
| 17 | "The A2A Proxy generates and serves an agent card descriptor" | The gateway returns the upstream agent's card with its advertised URLs rewritten; it does not synthesise a card | Not determined | **Fix** |
| 18 | Descriptor address `https://<gateway-host>/<context-path>/.well-known/agent-card.json` | Consistent with the path the gateway matches | The proxy Overview shows a gateway URL of the form `https://<host>/<context-path>` | **Confirmed** |
| 19 | Sample descriptor with `agentType: "AUTONOMOUS"`, `capabilities: { "mcp": true }`, and skills carrying only `name` and `description` | `agentType` is a property of a catalogued agent identity, not of an agent card. No `capabilities.mcp` field exists anywhere in the product. A skill on an agent card carries more than a name and a description | Not determined | **Contradicted — removed** |
| 20 | "An agent must have an identity before it can be exposed through the A2A Proxy" | n/a | Contradicted — a working proxy was created with no agent identity anywhere in the environment | **Fix** |
| 21 | Next steps link to the MCP policies page for the skill-authorization pattern | Depends on row 3 | An "Add policies to your A2A Proxy" page already exists as a sibling | **Fix** — link retargeted |
| 22 | Images | n/a | n/a | The page carries no figures, so there was nothing to verify or replace |

## Note for the writer (not a fix)

The page has no screenshot. "Create an A2A Proxy" is a four-step wizard the reader arrives at for the first time, which is one of the triggers in `style-guide/05-formatting-and-document-structure/images-and-figures.md`. Whether to add one is a writer's call, so the pipeline left the images alone.

## What could not be verified

There is no upstream A2A agent in this environment, so nothing that requires a real agent card fetch or a real delegation could be observed. Rows 1, 2, 17, and 19 are marked "not determined" on the live column for that reason, and their verdicts rest on the code cross-check alone. This is a partial verification, honestly reported.

## Other changes

- Added the missing `description:` frontmatter.
- Ran the style pass over the whole page. Every non-factual hunk was diffed and confirmed meaning-preserving.

### DIFF
--- a/docs/gamma/4.12/agent-management/build/expose-agent-with-a2a-proxy.md
+++ b/docs/gamma/4.12/agent-management/build/expose-agent-with-a2a-proxy.md
@@ -1,9 +1,9 @@
 ---
+hidden: false
+noIndex: false
 description: >-
   Expose an upstream agent behind the AI Gateway with an Agent-to-Agent (A2A)
   Proxy so that other agents can discover and call it.
-hidden: false
-noIndex: false
 ---
 
 # Expose your agent with the A2A Proxy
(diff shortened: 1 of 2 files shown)


paulatulis (approved):
Added a commit to edit a few things. 

- Removed "The upstream agent is never exposed directly to the network" — replaced with clients reaching the agent through the gateway rather than calling it directly.
- Scoped client authentication to the wizard's default plan ("The wizard secures the proxy with a default plan that uses either an API key or a client certificate") instead of stating it as a property of every A2A proxy.
- Narrowed the rewriting claim from "the URLs that the descriptor advertises" to "the agent's endpoint URL", and added that other URLs in the card, such as documentationUrl, pass through unchanged.
- Same fix in the descriptor section: "its advertised endpoint URL rewritten to the gateway address".
- Changed "shows the gateway URL that clients call" to "shows the configured gateway URL".
- Renamed the nav target in Next steps: agent identities live under Agent Identity, in the Catalog section.

Additions

- New ## Give a client access section covering: the proxy rejects unauthenticated requests including agent-card requests; create/open an application in the APIM console → subscribe to the default plan → approve the subscription (manual validation, so it starts pending) → copy the API key; send it as Authorization: Bearer <api-key>; X-Gravitee-Api-Key is not accepted; and a pointer to gateway-level TLS settings for mTLS.
- Third Next steps bullet linking to Authorization Management → A2A Agents, noting policies target an agent from the Catalog, so the agent must be imported first.
- Note in Step 1 that the wizard derives the context path from the name, so the suggested value can be accepted or replaced.

Cleanups

- Dropped the duplicated "authentication" from the Wire-level governance bullet (it now duplicates the client-authentication bullet above it).
- Moved description after hidden/noIndex in the frontmatter, matching the sibling pages.
- Split a long sentence and dropped "separately" — both Vale flags, now at 0 errors / 0 warnings / 0 suggestions.

tolaore (approved):

PR #2230, opened 2026-08-18
AGENT-DRAFT: AIAM-130 — Add the Cost Rate Limit policy

tolaore:
Adds a how-to for the **Cost Rate Limit** policy, and corrects statements on the **Token Rate Limit** page that the policy source and a running console both contradict.

## Why this changed shape

The original draft added settings to the existing Token Rate Limit page. The feature this PR was raised for ships a **separate policy** — `cost-ratelimit`, "Cost Rate Limit" — which caps consumer spend in dollars rather than tokens and had no coverage anywhere in the docs. That page is now the main deliverable.

The settings the draft added to the Token Rate Limit page are real, but several were wrong. They are corrected rather than reverted.

## Files

| File | Change |
| --- | --- |
| `docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md` | New page |
| `docs/gamma/4.13/agent-management/build/add-the-token-rate-limit-policy.md` | Corrections and style pass |
| `docs/gamma/4.13/SUMMARY.md` | Navigation entry for the new page |
| `docs/gamma/4.13/.gitbook/assets/gamma-aim-*.png` | One screenshot per page |

## How this was verified

Code truth is `gravitee-policy-token-ratelimit` at `1.1.3` — the schemas, `usage-ratelimit-core`, and both modules' generated reference.

Live truth is a Gamma console running both policy plugins at `1.1.3`, walked through the documented procedure end to end. Every setting below was read off the running configuration form or the schema the Management API serves to it.

Steps 6 and 7 of the procedure, **Save** and **Deploy**, were not exercised, so the "This deployable is out of sync" wording is still unverified. **Save** is visible in the screenshots.

## Verdict table

| Claim on the page | Code truth | Live truth | Verdict |
| --- | --- | --- | --- |
| **Time unit** is `SECONDS` or `MINUTES` | Enum is `SECONDS`, `MINUTES`, `HOURS`, `DAYS` | Dropdown offers Seconds, Minutes, Hours, Days | Contradicted — fixed |
| **Reservation amount (tokens)** defaults to `0` | No default; `minimum: 1`, and a strategy with an amount at or below `0` is rejected at validation | Served schema has no default and a minimum of `1` | Contradicted — fixed |
| `ASYNC_MODE` is "the fastest option, and the default" | Form default is `FALLBACK_PASS_THROUGH` | Form opens on "Count exactly - keep serving if the counter store is unreachable" | Contradicted — fixed |
| **Strategy** values differ on execution mode | Values differ on exact versus approximate counting | Labels read "Count exactly…" and "Count approximately…" | Contradicted — table rewritten with the label the console shows |
| A request that never reports usage "is still charged" | Token usage is fail-open; usage resolves to `0` and the reservation is released | Not separately exercised | Contradicted by code — fixed |
| `CALENDAR` aligns to the calendar hour, day, week, or month | Only a period of exactly 1 hour, 1 day, 7 days, or 30 days aligns; any other period is rejected at deployment | Label reads "Calendar-aligned window (hour, day, week, or month)" | Misleading — constraint added |
| On internal error, "headers report a counter of `0` and a reset of `-1`" | Sets `Remaining` to the full limit and `Reset` to `-1`, and only when **Add response headers** is enabled | Not separately exercised | Contradicted by code — fixed |
| Step 3: "In the **Request Phase** section, click **Browse all...**" | Not determined by code | The Request Phase has no such control. The plus opens a search, and **Browse full catalog** opens the **Add Policy** dialog. "Browse all…" belongs to the Response Phase | Contradicted — fixed |
| Step 4: "click **Add to flow**" | Not determined by code | Correct on the catalog path. The button is present on the policy's detail view | Confirmed — unchanged |
| Step 1: LLM Studio sits under **Design** | Not determined by code | The proxy nav has a **DESIGN** group containing Models and LLM Studio | Confirmed |
| Step 2: **Common Flows**, usually **Prompt** | Not determined by code | **COMMON FLOWS** lists Prompt, Embeddings, and Models | Confirmed |
| Defaults for **Reset strategy**, **Timezone**, **Reservation strategy**, **Reservation ceiling** | `ROLLING`, `UTC`, `NONE`, `0` | Form opens on those values | Confirmed — unchanged |
| Reservation is reconciled and refunded before the rejection; never exceeds the limit; forces a strict counter; a cached response counts as zero | Each confirmed in the engine | Not separately exercised | Confirmed — unchanged |
| `429` body text, and the three `X-Token-Rate-Limit-*` header names | Match the format string and header constants | Not separately exercised | Confirmed — unchanged |

Every setting on the new Cost Rate Limit page was read off the running form or the schema behind it, including the dollar fields, **Missing price policy**, and the per-model price fields it depends on.

## For the writer

1. **The screenshots come from a local test proxy** named "Verification Test Router". No customer, personal, or credential data is in either image, but retaking them against a better-named proxy would read better in published docs.
2. **Reservation ceiling must not be lower than reservation amount.** A configuration that breaks this is rejected at deployment. Neither page mentions it.
3. **The APIM policy reference for this policy is stale.** `docs/apim/*/create-and-configure-apis/apply-policies/policy-reference/ai-token-rate-limit.md` still lists **Time unit** as `SECONDS` or `MINUTES` and carries none of the newer settings, across every version that has the page. It is generated from the policy repository, so it needs a regeneration rather than a hand edit.
4. **The Gamma 4.12 copy of the Token Rate Limit page is untouched.** Whether it should change depends on which policy version 4.12 ships.
5. A `<!-- TODO: add an Overview -->` marker sits below the Token Rate Limit H1, per the structure rules. The new page has a written Overview instead. Drop the marker if the opening prose is considered sufficient.

## Style pass

Both files pass Vale with no errors. The one remaining warning is a single adverb. Meaning changes are confined to the rows marked *Contradicted* or *Misleading*; everything else is form only.


docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md
```diff
@@ -0,0 +1,147 @@
...
+## Settings
+
+### Cost budget
+
+The **Cost budget** section sets the amount, the window it runs over, and who it belongs to:
+
+| Setting | Description | Default |
+| --- | --- | --- |
+| **Max budget (dollars)** | The amount allowed per period, in dollars. Enter `5` for five dollars and `0.05` for five cents. Leave it at `0` to use the dynamic budget instead | - |
+| **Max budget (dynamic, dollars)** | An Expression Language expression that resolves to the budget, in dollars. Used only when the static budget is `0` | - |
```
vikrantgravitee: Setting only this and leaving the static budget at 0 fails to deploy with "A max budget is required". The deploy-time check reads the older micro-dollar field, not this one, so it never sees the expression. Same issue affects lines 45 and 59 - worth holding until the policy is fixed.

docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md
```diff
@@ -0,0 +1,147 @@
...
+| **Time duration** | How long before the budget resets | `1` |
+| **Time unit** | `MINUTES`, `HOURS`, or `DAYS` | `DAYS` |
+| **Key** | Identifies the consumer the budget applies to. Supports Expression Language. Leave it empty to count against the plan and subscription pair | Empty |
+| **Use key only** | Counts against the key alone, ignoring the subscription and plan. Ignored while **Key** is empty | `false` |
+| **Reset strategy** | `ROLLING` starts the window at the consumer's first request. `CALENDAR` aligns every consumer to the same clock boundary | `ROLLING` |
+| **Timezone** | The IANA identifier used for calendar-aligned boundaries, such as `UTC` or `Europe/Paris` | `UTC` |
+| **Reservation strategy** | `NONE`, `FIXED`, or `ADAPTIVE`. Decides whether the budget is checked against an estimate before the model is called | `NONE` |
+| **Reservation amount (dollars)** | The amount held per request while the model answers. Must be greater than `0` whenever a reservation strategy is set | - |
+| **Reservation ceiling (dollars)** | The largest amount `ADAPTIVE` may hold. Leave it empty for no ceiling | - |
```
vikrantgravitee: Also worth stating that a ceiling lower than the reservation amount is rejected at deploy, same as the rejections already documented on lines 54 and 59.

docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md
```diff
@@ -0,0 +1,147 @@
...
+
+Each dropdown shows a descriptive label rather than the stored value. **Time unit** reads **Minutes**, **Hours**, or **Days**. **Reset strategy** reads **Rolling window from first request** or **Calendar-aligned window**. **Reservation strategy** reads **No reservation**, **Reserve a fixed amount per request**, or **Reserve an amount learned from recent observed usage**.
+
+A budget of `0` with no dynamic budget is rejected when the API is deployed, so set one or the other.
+
+By default the budget belongs to a plan and subscription pair, not to a caller. When several identities share one subscription, they share one budget, and the first one to spend it exhausts it for everyone. Set **Key** to an expression that resolves the caller's identity, and enable **Use key only**, to give each identity its own budget.
+
+`CALENDAR` only aligns a window that matches a calendar period. Set **Time duration** and **Time unit** to exactly 1 hour, 1 day, 7 days, or 30 days. Any other period is rejected when the API is deployed, so use `ROLLING` instead.
```
vikrantgravitee: The rule is on the total length of the window rather than the unit, so 60 minutes is accepted as 1 hour. As written it reads like that would be rejected.

docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Reservations
+
+With no reservation, the policy checks the budget against spend already recorded. The real cost of a request is only known after its response has been delivered. One expensive call can therefore overshoot the budget, and concurrent calls can overshoot it together.
+
+Two **Reservation strategy** values hold budget up front. `FIXED` holds **Reservation amount** before the model is called. `ADAPTIVE` starts from that amount and learns from what the same consumer actually spends, up to **Reservation ceiling**. Either way the reservation is reconciled against the real cost once the response completes, and it's refunded before the rejection when the budget overflows. Estimation accuracy affects only how tightly the budget is held, never the recorded spend, because reconciliation applies the exact difference.
```
vikrantgravitee: "never the recorded spend" isn't quite true: when a model answers with no price and Missing price policy is FAIL_CLOSED, the charge is the larger of the reservation amount and the fallback cost, so the reservation does affect what is recorded. Suggest qualifying with "when the cost is known".

docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md
```diff
@@ -0,0 +1,147 @@
...
+
+A reservation never exceeds the resolved budget. Enabling a reservation also makes the counter strict, even under `ASYNC_MODE`. Adaptive estimates are held per gateway node and are lost on restart, so keep **Reservation amount** meaningful on its own.
```
vikrantgravitee: One more limit worth adding: each node only tracks 10,000 consumers, and past that it forgets older ones, which fall back to the plain reservation amount. That reinforces the advice already on this line.

vikrantgravitee (commented):
Checked the page against the policy implementation. Most of it matches. A few points below, one of which looks like a policy bug rather than a doc issue.

docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md
```diff
@@ -0,0 +1,147 @@
...
+| Count approximately - faster, but the limit can be overshot | `ASYNC_MODE` | Counts are buffered and flushed in the background. The fastest option and the lightest on the store, but concurrent requests can pass the budget before the flush catches up |
+
+Because a money budget is the kind of ceiling you don't want overshot, `FALLBACK_PASS_THROUGH` is the default rather than `ASYNC_MODE`.
+
+With `BLOCK_ON_INTERNAL_ERROR`, the rejection reports `Cost rate limit blocked the query due to internal error`. With `FALLBACK_PASS_THROUGH` and with `ASYNC_MODE`, the request is allowed. When **Add response headers** is enabled, `X-Cost-Rate-Limit-Remaining` then reports the full budget and `X-Cost-Rate-Limit-Reset` reports `-1`. If no rate-limit configuration is installed, the request fails with `No rate-limit config has been installed.`
+
+### Response headers
+
+Enable **Add response headers** to return the consumer's budget on every response:
```
vikrantgravitee: "Add response headers" doesn't appear in any settings table, so its default isn't stated anywhere. It is off by default, and every other setting on the page lists one.

docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md
garethbrinn: Held. Confirmed the cause: `CostRateLimitConfiguration` overrides `limitExpression()` to return `dynamicMaxBudgetUsd`, but `RateLimitConfigurationValidator.validateLimit` reads `getDynamicLimit()`, which is the inherited `dynamicLimit` field and stays null. So the dollar expression is never seen and the deploy fails with "A max budget is required". Removed the **Max budget (dynamic, dollars)** row, dropped "Leave it at `0` to use the dynamic budget instead" from **Max budget (dollars)**, and rewrote the prose as "A budget of `0` is rejected when the API is deployed, so set an amount greater than `0`." The field is left undocumented until the policy is fixed. The Token Rate Limit page keeps its **Max tokens (dynamic)** row: that one writes `rate.dynamicLimit`, the field the validator actually reads.

docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md
garethbrinn: Added. **Reservation ceiling (dollars)** now reads "Must not be lower than **Reservation amount**". Matches `validateReservation`, which rejects a ceiling greater than `0` and lower than the amount. Added the same to the Token Rate Limit page, which had no mention of it either.

docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md
garethbrinn: Fixed. `supportsCalendarAlignment` converts the period to minutes before matching, so the rule is on the total length. Rewritten as: "`CALENDAR` only aligns a window whose total length is a calendar period: an hour, a day, 7 days, or 30 days. The check is on the length of the window rather than on the unit you express it in, so `60` **Minutes** is accepted as an hour, and `24` **Hours** as a day." Same correction applied to the Token Rate Limit page, which carried the same wording.

docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md
garethbrinn: Qualified. `UsageRateLimitEngine` settles an unmeasured request under `FAIL_CLOSED` as `Math.max(reservation.amount(), missingUsageFallback())`, so you are right. The sentence now reads "When the cost is known, estimation accuracy affects only how tightly the budget is held, never the recorded spend", followed by "When the cost is unknown, the reservation can set the recorded spend" and a link to Missing price policy. That section now states the larger-of rule explicitly.

docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md
garethbrinn: Added. `AdaptiveUsageEstimator.DEFAULT_MAX_TRACKED_KEYS` is `10_000` and `observe` evicts an existing entry once the map is full. Written as: "Each node tracks at most 10,000 consumers, and past that it drops an entry it already holds. The dropped consumer reserves the plain **Reservation amount** until the node observes it again." One wording note: eviction takes the first entry the key iterator yields, which for a `ConcurrentHashMap` isn't insertion-ordered, so the page says "an entry it already holds" rather than "the oldest". Added to the Token Rate Limit page too, which shares the estimator.

docs/gamma/4.13/agent-management/build/add-the-cost-rate-limit-policy.md
garethbrinn: Added. Confirmed `addHeaders` has `"default": false` in both schemas. The Response headers section now opens with a settings table giving **Add response headers** its own row and a default of `false`, above the header list. Same on the Token Rate Limit page.

garethbrinn (approved):

PR #2233, opened 2026-08-20
Add: connect and secure the Atlassian and Stripe MCP servers

samprodger:
Adds two how-to guides to the Agent Management section, alongside the existing GitHub MCP server guide:

- Connect and secure the Atlassian MCP server
- Connect and secure the Stripe MCP server

Each covers catalog registration, tool curation in MCP Studio, OAuth2 authentication, per-tool authorization with GAPL, rate limiting, PII redaction, and observability, and follows the structure of the GitHub guide so the three read as a set.

Both pages are added to the 4.12 and 4.13 spaces with their screenshots and `SUMMARY.md` entries.

### Notes for review

Both guides were written from a working build, so a few points are specific rather than generic:

- **Atlassian publishes several MCP endpoints and they expose different tools.** The guide names the one that serves the Jira and Confluence tools it composes, and flags that it is an early-access endpoint subject to change.
- **Tool identifiers are lowercased before policy evaluation.** Atlassian's tool names are camelCase, so the policy examples use the lowercased form. GitHub's are already snake_case, which is why the existing guide doesn't hit this.
- **A mistyped resource in a `forbid` statement fails open**, where the same mistake in a `permit` fails closed. Both guides say so and tell the reader to confirm a denial rather than read the policy back.
- **Whether the two logged response legs differ depends on the upstream's response transport.** Where the upstream streams, the endpoint leg records the payload before response policies ran, so redaction alone doesn't keep personal data out of the log store. The Atlassian guide states this, and the Stripe guide covers the JSON case.

### One thing worth a second pair of eyes

The registration wizard's submit button is written as **Import MCP Server**, matching the existing pages. Reading the Agent Management module bundle, the only occurrence of that phrasing is an error string, and the submit control appears to be labelled **Save catalog entry**. I could not settle it confidently, so I have kept the existing wording for consistency rather than change it here. If someone can confirm against the console, it may be worth a small sweep, since the same label appears on the GitHub guide and on `register-an-mcp-server.md`.


garethbrinn:
@samprodger the verification tool rasied the following questions:
1. The Stripe compose screenshot shows stripe-mcp with only 4 tools discovered, all 4 selected — stripe_api_write isn't there. Was it absent because the discovery key lacked write scope, or is it gated another way?

2. The Atlassian compose screenshot shows atlassian-mcp-server with only 5 tools, all 5 selected — execute and discover aren't there. Does /v1/mcp/preview ever expose them, and under what conditions?

3. Does the policy engine evaluate MCP tool identifiers in lowercase (atlassian-mcp-server.getjiraissue) while the console and tools/list show original casing? If so, does the same apply to Stripe tool names?

4. In Gamma's Policy Studio, does the Rate Limit policy render as Limit / Period / Add headers, or with the raw schema titles (Max requests (static), Static time duration, Add response headers)?

5. Where does "the Gateway picks up a deployed policy within 30 seconds" come from, and is that interval configurable?

6. Are a plan's required scopes written from the composition at creation and then left unchanged, so adding a tool later leaves the plan stale? And what casing do those scopes use in the WWW-Authenticate header?

7. Are agent-management and authorization-management the exact licence pack identifiers, and does a missing pack really render as a locked upgrade prompt rather than an error?

8. Is the logging difference between Atlassian and Stripe (streaming logs the endpoint leg pre-redaction; JSON logs both legs redacted) a transport-driven rule a reader can predict, or incidental to these two servers?

9. Does an Atlassian API token have to be created against the Rovo MCP V2 application, and do tokens from another application really authenticate and complete discovery but fail every call?

10. Does getAccessibleAtlassianResources sit in a different scope family from the Jira tools, so a token that calls Jira tools may not be able to list resources?

11. Does Atlassian's edge reject some default User-Agent values with a 403 that looks like an authorization failure?

12. Does Stripe manage MCP access independently for sandbox and live mode?

13. Is Stripe's finest permission grouping really charges-and-refunds combined, so a key that can refund can also write charges?

14. Do Stripe's write tools accept an optional human_confirmation argument that the agent itself decides whether to invoke?

samprodger:
@garethbrinn thanks, that tool earned its keep. Two of these were real hits on the PR and both are now fixed. I went to the running console and to the gateway's log store rather than answering from memory.

Pushed in 93727026d.

Tagging each answer with how it is known: **measured** on the 4.12.15 build, **documented** by the
vendor, **derived** from source or schema without a console look, or **unverified**.

---

**1 and 2, the compose screenshots. Confirmed in the console, and you have found a real problem
with the images rather than with the build.** Both were taken with the **Selected only** filter
active. I opened the same screen on the running stack with the filter off, and it reads:

```
26 visible · 5 selected
  atlassian-mcp-server    17 tools   5 selected
  stripe-mcp               9 tools
```

So the catalog holds all 17 Atlassian tools and all 9 Stripe tools, and the studio selects 5. The
committed screenshots show only what was kept and hide precisely the tools whose absence is the
entire argument. **Both need re-shooting with the filter off**, so they show 9 to 4 and 17 to 5
rather than 4 to 4 and 5 to 5.

To answer what you actually asked:

- **Stripe, measured.** Discovery returns **9** tools and `stripe_api_write` is one of them. It was
  not absent and it was not gated: it was deliberately not composed. Worth stating plainly because
  it is the finding the whole page rests on: the tool list **does not vary by key type**. An
  unrestricted `sk_test_` key and an `rk_test_` restricted to Charges and Refunds both returned the
  same nine tools, `stripe_api_write` included. Key scope is enforced at call time, not at
  discovery.
- **Atlassian, measured.** `/v1/mcp/preview` does expose `discover` and `execute`, in the 17 that
  discovery returns. Both were deliberately withheld. `execute` is the one that matters: it is a
  generic executor that reaches whatever `discover` can find, so composing it would make the
  allowlist meaningless.

**3, tool identifier casing. Measured, and yes.** The PEP evaluates
`resource=MCPTool::"atlassian-mcp-server.getjiraissue"` while the console, the Catalog and
`tools/list` all show `getJiraIssue`. The same lowercasing applies to Stripe, but you cannot see it,
because Stripe's tool names are already lowercase: the decision log reads
`stripe-mcp.stripe_api_read`, identical to the original. That is exactly why a GitHub-only or
Stripe-only guide never surfaces the trap and an Atlassian one always does.

**4, Rate Limit field labels. Verified in the console, and your tool is right.** I opened the
policy on the running `atlassian-triage-toolbelt` studio. The form renders the raw schema titles:

| Our page said | The console actually says |
|---|---|
| **Limit** | **Max requests (static)**, set to `5` |
| **Period** | **Static time duration** `60` and **Static time unit** `SECONDS`, two separate fields |
| **Add headers** | **Add response headers** |
| **Key** | **Key**, correct as written |
| **Error strategy** | **Error strategy**, correct as written |

The Period one was worse than wording: there is no single Period field, so a reader following that
step looked for one control and found two. **Fixed on both pages.**

**5, the 30 second figure. Unverified, and our own measurement disagrees.** It is inherited from the
published GitHub page and I cannot trace it further than that. This build measured roughly **40
seconds**, watching the PDP engine log each commit as `AuthzPdpEngine commit g8 -> g9`. Whether the
interval is configurable, I do not know. I would replace the number with the measured one or drop
the number and say a deployed policy is picked up without a restart.

**6, plan scopes. Half measured, half untested.** The casing half is measured and worth putting in
the page: the advertised scopes use the **original** casing, `searchJiraIssuesUsingJql getJiraIssue
transitionJiraIssue searchConfluence getConfluenceContent`, captured verbatim from the
`WWW-Authenticate` header on an unauthenticated call. So scopes keep camelCase while policy resource
identifiers are lowercased, which is a trap worth naming since both strings name the same tool.
Whether recomposing a studio rewrites the plan's scopes or leaves them stale was never tested.

**7, licence packs. Reported from the build, worth one confirmation.** The packs are named
`agent-management` and `authorization-management`, and a universe-tier licence missing them rendered
the module as a locked upgrade prompt rather than an error, which reads as "not in this version".
That came out of the build rather than from a deliberate negative test, so I would not print it
without someone reproducing it.

**8, the logging difference. Measured on both servers, and it is real.** I read the gateway's own
log store rather than reasoning about it: `gravitee-v4-log-2026.08.13` in Elasticsearch, both
response legs of the original calls.

| Server | endpoint-response leg | entrypoint-response leg |
|---|---|---|
| Atlassian, permitted `getJiraIssue` | 520 bytes, email address **in clear**, no redaction markers | 496 bytes, 4 x `[REDACTED]` |
| Stripe, permitted `stripe_api_read` | 1560 bytes, 8 x `[REDACTED]` | 1560 bytes, 8 x `[REDACTED]` |

Eight Atlassian calls show the same split, so it is not a one-off. On Atlassian the endpoint leg is
recorded before response policies run, which means **personal data lands in the log store in clear
even though the agent receives `[REDACTED]`**. On Stripe both legs are recorded after redaction.

To answer your question directly: the streaming explanation fits, since Atlassian streams and Stripe
returns a single JSON body, but two upstreams is not enough to publish as a rule a reader can
predict. Both pages now state their own measured behaviour and tell the reader to confirm theirs
with one permitted call.

Worth saying plainly because it is the part that bites: payload logging is its own decision, not a
consequence of redaction. A redacted response does not imply a redacted log store.

I also owe you a correction. My fact sheet had carried the Stripe result across to Atlassian as
though it had been measured there, and on the strength of that I briefly removed the correct
statement from the Atlassian page. The page was right the first time. It is restored, and now cites
the measured byte counts.

**9, the token application. Measured, and yes to both halves.** Tokens from the wrong application
authenticate, open a session and return the full tool list, then fail on the first real call:

| Token application | `atlassianUserInfo` | `searchJiraIssuesUsingJql` |
|---|---|---|
| Rovo MCP V2 | 403, needs `read:me` | 200, real issues |
| `appId=mcp` | 200 | 403, `Insufficient scopes ... [search:jira:agent-interface]` |

The scope families are disjoint, and scopes cannot be edited after a token is created, so the
application chosen at mint time decides whether the integration works. There is no single token
covering both families.

**10, `getAccessibleAtlassianResources`. Inferred, not tested.** The disjoint-families result above
was measured with `atlassianUserInfo`. That `getAccessibleAtlassianResources` sits in the same
identity family was assumed from its shape, not proven. I would soften that hint or test it.

**11, the User-Agent rejection. Measured, and yes.** A permitted call returned
`403 error 1010 browser_signature_banned` from Atlassian's edge, which is indistinguishable from an
authorization failure at the client. Fixed with a Transform Headers policy on the `tools/call`
request phase setting an explicit `User-Agent`.

**12, Stripe sandbox and live. Documented, not observed.** Stripe's own documentation states MCP
access is enabled per environment at `dashboard.stripe.com/settings/mcp` and managed separately for
sandbox and live. This account had it enabled, so the disabled state was never seen, and Stripe does
not document what a caller gets when it is off. The page deliberately does not guess that error.

**13, charges and refunds. Yes, by construction.** The restricted key used here was granted write on
**Charges and Refunds** as a single resource, because Stripe's permission model has no finer split.
A key that can refund can also write charges.

**14, `human_confirmation`. Measured from the live tool schema, and yes.** Both write tools accept an
optional `human_confirmation` object whose description instructs the agent to surface an approval URL
to a human and re-invoke with an `approval_token`. It appears nowhere on Stripe's MCP documentation
page. It is a real control and the page says so, with the caveat that it is optional and the agent
decides whether to use it, so the agent most likely to skip it is the one that has been talked into
making the call.

---

## Already changed in the PR

1. **Rate Limit step rewritten** to the real console labels, on both pages, including splitting the
   window into its two actual fields.
2. **The 30 second figure is gone.** Both pages now say a deployed policy is picked up without a
   restart, typically within a minute, which is what this build measured.
3. **Console labels corrected** where our own committed screenshot disagreed with the prose:
   **Test connection** for "Verify URL", **Static credentials** for "Static credential", **Save
   catalog entry** for "Import MCP Server".
4. **Example server names corrected** to `atlassian-triage-toolbelt` and `finance-support-toolbelt`,
   matching the console and the screenshots, in prose and in the sample requests.
5. **The Atlassian logging claim is no longer asserted.** The page previously stated that streaming
   puts the endpoint leg pre-redaction and told the reader to expect an email address in clear.
   Verification now confirms redaction by comparing a direct upstream read against a read through
   the Composite MCP Server, which holds whichever way the logging turns out, and the reader is told
   to check what their own log store recorded. The Stripe page keeps its measured result but no
   longer generalises it into a transport rule.
6. **`getAccessibleAtlassianResources`** softened to what was actually tested.
7. **The plan-scope staleness claim softened.** The page asserted that scopes are written at
   creation and stay as set, which is your question 6 and was never tested.

8. **Both compose screenshots re-shot** on the running stack with the filter off, and installed in
   both versions. The Atlassian one now shows all 17 discovered tools with the 5 composed ones
   ticked and `discover` and `execute` visibly unticked. The Stripe one shows all 9 with
   `stripe_api_write` visibly unticked, which is the single most useful frame in that page. Both
   captions updated to describe what is now on screen. The wizard chrome is cropped out, because
   the edit flow shows two steps where the create flow shows five and I did not want a new
   image-versus-prose mismatch.

9. **The logging behaviour is settled and both pages now cite measured figures**, per question 8
   above. No token needed in the end: the original calls were still in the log store, so the answer
   came from the recorded documents rather than from new traffic.

## Nothing outstanding

Every question has an answer, and everything that was an edit has been made.

## Separately, from auditing the screenshots

The compose shots show `atlassian-triage-toolbelt` and `finance-support-toolbelt` while the prose
says `-toolset` six times per page, and the add-server shot shows the console saying **Static
credentials**, **Test connection** and **Review catalog entry** against prose that says "Static
credential", "Verify URL" and "Import MCP Server". The images and the text disagree inside the same
PR, which also means the published GitHub page carries the same four label errors.

samprodger:
@garethbrinn one more, and this one is a correctness bug I shipped rather than a question from your tool. Found by re-verifying every claim against the running stack instead of against my own notes. Pushed in fca01d5e8.

**Both pages published an authorization policy that cannot work on the setup the same pages tell you to build.**

They showed:

```
permit ( principal in Group::"<delivery-group-id>", ... )
```

Enabling FGA sets the Authorization PEP's subject type to `User`, so the Policy Decision Point receives the caller as `User::"<id>"`. I checked the Authorization Management entity store on the build behind these guides: **0 entities, empty schema**. A `principal in Group::"..."` statement therefore matches nothing.

The consequence is worse than a non-working example, because it fails in both directions:

- the `permit` matches nothing, so every call is denied with `NOT_APPLICABLE / "No applicable policy"`
- the `forbid` also matches nothing, so it silently never fires, and any broader permit still applies

A reader following the guide exactly would have had the whole thing deny everything, then, on widening the permit to get moving, would have had a `forbid` that looked deployed and did nothing.

The five policies actually deployed on that build all use `principal == User::"<id>"`. Both pages now publish that form, plus a note saying where to get the identifier (the decision log records `subject=User::"..."` on every evaluated call) and what group principals additionally need: the identity provider's groups synced in as entities and the subject type changed to match.

I have corrected the same error in my own fact sheets, which had claimed the group form was verified when it never was.

**Also in this push:** the PII Filtering step now names the real fields, **AI Resource** and **PII Categories**, and says that the policy sits in the request phase and still filters the response payload, which is not obvious and is what actually redacts the tool result.

**What else I re-verified against the live stack, all of it now matching the pages:** the composed tool sets (5 and 4, exact names from the deployed API definitions), the discovered counts (17 and 9), the `WWW-Authenticate` scope headers on both servers including `stripe_api_write` being absent, the catalog entity IDs (`mcp-server.atlassian-mcp-server`, `mcp-server.stripe-mcp`), the lowercased action in the decision log (`atlassian-mcp-server.getjiraissue`), the full authz-pep config, the `User-Agent: curl/8.7.1` transform, the rate limit values (5 per 60 SECONDS, `BLOCK_ON_INTERNAL_ERROR`, key on `user`, `useKeyOnly`), and every Rate Limit and PII Filtering field title against the shipped plugin schemas.

garethbrinn (approved):

PR #2248, opened 2026-08-24
OBS-37: Document the Tenant filter on environment-level Logs

tolaore:
https://gravitee.atlassian.net/browse/OBS-37


tolaore:
Screenshot also attached

tolaore:
Re-verification of the removed filter-section content, at the writer's request — where each removed element actually lives in the console source at tag 4.12.15:

- Quick filters (Period), Entrypoints filter, Reset filters, More-filters form: they exist on the API-level v4 runtime-logs screens (api-traffic-v4/runtime-logs/components/api-runtime-logs-quick-filters.component.html:21-31,117 and the more-filters form), which view-api-logs docs cover — not on the environment Observability Logs screen.
- Reset filters additionally appears in gio-metadata, API Products subscriptions, and the API analytics filter bars. Nowhere in the observability env-logs module.
- 'Additional filters' exists only as an HTML comment in the legacy alerts page. 'Error Types' appears nowhere in console src. 'Response time (ms)' only as chart axis names.
- The environment Logs template is identical at every 4.12 tag checked (4.12.0, 4.12.5, 4.12.10, 4.12.13, 4.12.14, 4.12.15) and on master: gd-timeframe-selector + gd-dynamic-filter-bar only. The removed block never matched this screen at any point in the 4.12 line.
- At 4.11.25 the Observability module had no logs screen at all (dashboards only); the era's environment-wide logs were the legacy Analytics platform-logs screen, whose top-level template contains none of these strings either (deeper i18n layers not chased — it is not the screen this page's navigation points to).

Conclusion: the removed content describes other screens (primarily the API-level runtime-logs UI), which have their own pages. The rewritten section matches the screen this page documents.

tolaore:
Attached two writer-captured figures (claims table rows 31-32): the Add a filter dialog with the Filter by list open now illustrates the 'filter the logs by the following criteria' list (new figure), and the writer's full Logs screen replaces the synthetic overview capture (same filename, content swap). Both were captured on an internal development console, so all internal data is blurred before the pixels ship: the environment name, API names and paths, gateway pod hostnames, and the account avatar and notification dot. Redaction is the toolkit's pixelate-then-blur with a zoom-crop ghost-text check on every region. One caveat for review: the source console runs a newer build, so its sidebar shows a couple of items (API Score, Portal Settings) that a 4.12.15 Console doesn't have - the filter dialog and Logs table themselves are identical to 4.12.15 (template diff between origin/4.12.x and master is empty for this module). The Tenant dialog and applied-filter figures are unchanged.

SamHardi (approved):

garethbrinn (approved):

PR #2249, opened 2026-08-24
docs(apim): document http.pathHandling on 4.11, 4.12 and 4.13

ytvnr:
**Issue**

https://gravitee.atlassian.net/browse/APIM-14917

**Description**

New documentation for the `http.pathHandling` Gateway setting, under Configure and Manage the Platform > Gravitee Gateway, on 4.11, 4.12 and 4.13. The setting decides how the Gateway treats the request path before it resolves the listener context path, and therefore before it enforces any plan. It ships in 4.11.26, 4.12.18 and 4.13.0, from gravitee-io/gravitee-api-management#19166, gravitee-io/gravitee-api-management#19387 and gravitee-io/gravitee-api-management#19388.

- **4.11 and 4.12 are byte-identical.** The default is `RAW`, so upgrading to 4.11.26 or 4.12.18 changes no behavior by itself, and closing the exposure stays an explicit operator decision. The two branches differ in how the normalizer is carried, but none of that is visible to a user, so the pages say the same thing.
- **4.13 leads with the breaking change**, and is a folder of six pages rather than a single one: the overview, the upgrade to the new default, how to configure the mode, how to verify it on a running Gateway, the reference, and the limits. The default becomes `NORMALIZE`: what it changes for a request, the decoding of percent-encoded unreserved characters, how to roll back to `RAW` and what that restores, then a short upgrade checklist.
- All three versions cover what the modes do **not** cover (`..%2f..`, `%252e%252e`, `%c0%ae`, `%00`, backslash separators) and state plainly that `REJECT` is not traversal hardening for an arbitrary receiver, the deliberate departure from RFC 3986 on `..;x`, that `uri` keeps reporting the bytes as received in every mode, and that the setting is Gateway-wide.
- Also documents the companion flag `handlers.rejected.analytics.enabled`, which defaults to `true` where the not-found equivalent defaults to `false`.

Companion changes: a pointer from **Configure your HTTP Server**, a `SUMMARY.md` entry per version, a 4.13 entry in **Breaking Changes and Deprecations**, and a highlight plus a breaking-change section in the 4.13 release notes.

**Additional context**

The 4.13 pages are registered in `docs/apim/4.13/.version-overrides`, since they diverge from 4.12 on purpose and the sync bot would otherwise overwrite them. `prepare-a-production-environment/configure-your-http-server.md` is registered there too: its cross-reference points at the flat page on 4.12 and at the folder README on 4.13, and without that entry the first later edit of the 4.12 page would restore a link to a page that doesn't exist on 4.13.

Two items from the ADR are deliberately absent: the `RAW` plus ingress escape hatch, which the ADR marks as untested, and the recommendation to enable `NORMALIZE` on the production best practices page, which belongs in its own PR.


garethbrinn:
Hi @ytvnr 

Here are some questions about the PR from the verification tool:
1. *Which patch releases carry this, on 4.11, 4.12 and 4.13?* #19166 is still open against 4.11.x and `pathHandling` is on no release branch — a released 4.12.15 gateway has no such key, class or startup line. The 4.11 and 4.12 pages currently read as if it's already shipped.

2. *Is NORMALIZE-by-default committed to 4.13.0, or intent for now?* Your branch binds `${http.pathHandling:RAW}` and the PR describes the flip as a master follow-up. Three things in the docs PR state it as fact — the 4.13 page, the breaking-change entry and the release-note highlight.

3. *Can you confirm the malformed-percent sentence is wrong for RAW?* The page says `/proxyv4/a%+41b` answers 400 in every mode including RAW, from the HTTP layer before the Gateway sees it. On 4.12.15 the Gateway routed `a%+41b`, `a%zz` and `a%` and forwarded all three upstream — the 400 came from the backend. Should it say that under RAW the Gateway neither refuses nor rewrites it, so the 400 depends on the receiver?

4. *Does consuming the normalizer from a released gravitee-common on 4.12.x/master change any behavior vs the vendored 4.11 copy* — the `..;` strictness especially? The 4.11 and 4.12 pages are currently byte-identical, so any divergence breaks that.

5. *gRPC — does the refusal actually reach a client as trailers?* The page explains `grpc-status: 3` that way; the code sets response headers keyed off the request's `Content-Type: application/grpc`. Does it hold on HTTP/1.1, or only HTTP/2?

6. *Will `handlers.rejected.analytics.enabled` get a real values.yaml entry before GA?* It works today only because the chart passes `gateway.handlers` through verbatim, and the PR lists surfacing it as out of scope.

7. *Should the 4.11/4.12 pages carry the bypass narrative at all,* or document the three modes neutrally and leave the "what you were exposed to" framing to the 4.13 breaking change?

8. *Is anything in the ADR referenceable publicly,* or is it Confluence-internal only?

ytvnr:
Hi @garethbrinn, thanks for these. Several of them are real defects, so I'm going through them one at a time and will answer the rest in a follow-up. Starting with the first, since it changes the status of this PR.

**1. Which patch releases carry this, on 4.11, 4.12 and 4.13?**

You're right, and the honest answer is that we don't know yet. #19166 is still open against `4.11.x` and is merged nowhere, so `pathHandling` is on no release branch. The latest published patches are 4.11.25 and 4.12.17, both from 21 August, and neither carries the key, the class or the startup line.

The pages as written say nothing about this. They describe the setting in the present tense, which reads as "it's there". The failure case that worries me isn't a reader who can't find a key: it's a reader on 4.12.17 who sets `http.pathHandling: REJECT`, restarts, sees nothing in the logs and concludes the exposure is closed. It isn't, and the startup line the page offers as the way to confirm the active mode doesn't exist on that binary either. A false sense of protection on an authorization bypass is worse than a missing key.

So **this PR is now a draft**, and it shouldn't be merged until the availability versions are pinned. What we'd like your eyes on meanwhile is the substance: whether the three pages say the right things, in the right order, with the right warnings.

The TODO before it can leave draft:

- pin the patch releases for 4.11.x and 4.12.x, once the backports are decided
- confirm whether the 4.13 default lands in 4.13.0 or later, since the 4.13 page states "from 4.13.0" in four places
- add the availability statement to all three pages, reusing the idiom already present in `configure-your-http-server.md` ("Starting with Gravitee APIM 4.10, the Gateway can reject mTLS connections..."), so it's stated once, at the top

Answers to the other seven points to follow.

ytvnr:
Hi @garethbrinn, thanks for these. Answers to all eight below. Three were real defects and are fixed on the branch; one of my own earlier answers was wrong and is corrected here.

**1. Which patch releases carry this, on 4.11, 4.12 and 4.13?**

4.11.26, 4.12.18 and 4.13.0, now stated at the top of each page. When you asked, the answer was that we didn't know: nothing was merged anywhere. That's since changed, #19166 on `4.11.x`, #19387 on `4.12.x` and #19388 on `master` were all merged on 26 August.

The failure case behind your question was worth closing properly. It isn't a reader who can't find a key, it's a reader on 4.12.17 who sets `http.pathHandling: REJECT`, restarts, sees nothing in the logs and concludes the exposure is closed. So the pages now also say that upgrading to 4.11.26 or 4.12.18 changes no behavior by itself: the default stays `RAW` until someone sets `REJECT` or `NORMALIZE` explicitly.

**2. Is NORMALIZE-by-default committed to 4.13.0, or intent for now?**

Committed, and now shipped in #19388. Your reading of the code was right at the time, the flip lived in a master follow-up that hadn't been opened yet. No wording change was needed.

**3. Can you confirm the malformed-percent sentence is wrong for RAW?**

You're right, and it's corrected. That sentence came from a manual run on a local 4.11 gateway and I generalised it to three versions without re-checking, which is the kind of claim that can't survive a Vert.x major (4.5.27 on 4.11, 5.0.12 on 4.12 and master). Your 4.12.15 observation and the 4.11 one can both be true.

The pages now say: under `REJECT` and `NORMALIZE` a malformed percent sequence is answered with `400` because it has no normalized form. Under `RAW` the Gateway neither refuses nor rewrites it, so what the caller receives depends on the rest of the chain and can differ between versions.

**4. Does consuming the normalizer from gravitee-common change any behavior versus the vendored 4.11 copy?**

I answered this once already and got it wrong, so to be explicit: I claimed there was no 4.12.x PR and no master PR. Both existed when I wrote that. My search was scoped to the ticket key and didn't match them.

Your concern was legitimate. #19387 does consume `RequestPathNormalizer` from `gravitee-common` rather than vendoring it into the gateway reactor.

What it doesn't change is the byte-identity of the 4.11 and 4.12 pages, which is what the question was really about. #19387 cherry-picks the twelve commits from 4.11.x unchanged and adds two on top: the switch to `gravitee-common`, and an adaptation to the Vert.x 5 APIs where `RejectedPathProcessor` moves from `io.grpc` to `io.vertx.grpc.common.GrpcStatus`, with the value put on the wire unchanged. Nothing user-visible diverges between the two branches, `..;` strictness included, so the two pages can stay identical.

**5. gRPC, does the refusal actually reach a client as trailers?**

Good catch, the page was wrong. The code sets response *headers*, not trailers. The explanation is gone, and the page now describes only what the Gateway does: when the request's `Content-Type` is `application/grpc`, it sets `grpc-status: 3` (`INVALID_ARGUMENT`) and `grpc-message` on the response.

Your HTTP/1.1 versus HTTP/2 question is a real one but belongs on the code PR: the class comment in `RejectedPathProcessor` says the caller reads the status from the trailers, while the code sets headers, so the two disagree.

**6. Will `handlers.rejected.analytics.enabled` get a real values.yaml entry before GA?**

A chart question rather than a doc one, and not blocking here. The documented form works today and not by accident, the gateway configmap renders `gateway.handlers` verbatim. A first-class key would be easier to discover but wouldn't change what the page says.

**7. Should the 4.11/4.12 pages carry the bypass narrative at all?**

Keeping it, after weighing your point seriously. The case for removing it is that we'd be publishing a working reproduction against the default of every deployed 4.11 and 4.12 gateway. What settles it is that we already have: the same table is in the description of #19166, on a public repository, since the day it was opened. Taking it out of the docs buys nothing and costs an operator the reason to act.

Your question did surface a real gap, now closed, and it's the one described under point 1: the pages never said that upgrading alone changes nothing.

**8. Is anything in the ADR referenceable publicly?**

No, and nothing links to it. The three pages carry no Atlassian, Confluence or GitHub link and no ticket reference. Their only outbound links are to RFC 3986. The ADR lives in an internal space, is still marked PROPOSED, and names the customer behind the escalation.

---

**One more error, found while checking the above.** The 4.13 page said an unrecognized `http.pathHandling` value falls back to `RAW`, and that a typo therefore drops a 4.13 deployment back to the pre-4.13 behavior. That's true on 4.11.x and 4.12.x and the opposite of what master does: the fallback there is `NORMALIZE`, deliberately, so that a typo in a configuration file can't reopen the bypass. The page presented a protective design as a footgun. Corrected, with the maintenance-branch behavior stated next to it.

Same root cause as point 3, extrapolating one branch's code to another. Now that all three branches are merged, the remaining behavioral claims get re-checked against the real code before this leaves draft.

ytvnr:
@garethbrinn the split reads well, and I checked the mechanics rather than just the prose: all six pages are in `SUMMARY.md` with no orphans, no broken links across the folder or the three companion files, and the version lint passes (replayed locally against the branch head). Keeping the old flat `request-path-handling.md` entry in `.version-overrides` was the right call, `prune_overrides.py` treats it as a tombstone precisely so the sync bot cannot recreate it next to the folder.

Two things to fix, one of which will bite silently later.

**1. `configure-your-http-server.md` needs an override entry**

The cross-reference now differs between versions:

```
4.12: ../configure-and-manage-the-platform/gravitee-gateway/request-path-handling.md
4.13: ../configure-and-manage-the-platform/gravitee-gateway/request-path-handling/README.md
```

and that file is not listed in `docs/apim/4.13/.version-overrides`.

The lint passes here only because both versions happen to be modified in this PR. It checks "changed on both sides", not "identical on both sides", so it cannot see a permanent divergence introduced without an override.

After merge, the first edit to the 4.12 page (a busy one, it covers TLS, keystores, CRL and timeouts) triggers the sync bot, which copies it over the 4.13 one and restores a link to `request-path-handling.md`, which no longer exists on that version. That would be a broken link on the 4.13 docs, introduced by an auto-merged bot PR, with nothing pointing back at this change.

One line fixes it:

```
prepare-a-production-environment/configure-your-http-server.md
```

**2. The README's Next steps lists three of the five sub-pages**

"Verify Path Handling on a Running Gateway" and "Request Path Handling Reference" are missing from it. Both are in `SUMMARY.md` so they are reachable from the sidebar, but a reader navigating from the page body will not find them.

**One thing worth being deliberate about**, not a request. 4.13 is now a folder of six pages while 4.11 and 4.12 stay a single page. Any future change that applies to all three versions has to be written once on the flat page and then placed into one of six on 4.13, which is the work the sync bot normally absorbs for us. Given that the 4.13 page carried an upgrade narrative the other two don't, and ran past 300 lines, the split makes sense to me. I am only flagging that the maintenance asymmetry is now permanent for this topic.

ytvnr:
@garethbrinn both points are addressed, and I checked them on the branch head rather than in the diff alone.

**1. Override entry.** `prepare-a-production-environment/configure-your-http-server.md` is now in `docs/apim/4.13/.version-overrides`. The two comment lines above it are ignored by `sync_versions.py`, `prune_overrides.py` and `lint_overrides.py`, so they document the reason at no cost. The divergence they describe is still real on the branch: 4.12 links the flat page, 4.13 links the folder README.

**2. Next steps.** The README now lists all five sub-pages, in the same order as `SUMMARY.md`, and every target exists in the folder.

Nothing else changed in the pages since my last pass. The rest of the diff comes from the merge with `main`, on the environment logs and the next-gen cloud IP ranges.

I also refreshed the PR description, which still claimed `gravitee-io/gravitee-api-management#19166` was open and that the availability versions would be pinned before merge. All three gateway PRs merged on 26 August and the versions are stated in the pages.

So this is good to go for me. Approve and merge whenever it suits you, I can't approve my own PR.

garethbrinn (approved):

PR #2252, opened 2026-08-25
docs(apim): document connection-closed/reset retry for idempotent requests

carlos-andres-osorio:
## Issue

[APIM-14873](https://gravitee.atlassian.net/browse/APIM-14873)

## Purpose

Under sustained load, a pooled keep-alive connection can be closed or reset by the backend or an intermediary (load balancer, OpenShift Route) in the window between the Gateway taking it from the pool and writing the next request to it. Before 4.12.18, this race surfaced directly as a client-visible `502` (`GATEWAY_CLIENT_CONNECTION_CLOSED` / `GATEWAY_CLIENT_CONNECTION_ERROR`) with no built-in retry, forcing customers to work around it with Failover/Retry policy configuration. From 4.12.18, the Gateway retries such a request once on a fresh connection when it is safe to do so (body-less and idempotent). This PR documents that new behavior.

## Summary of changes

- `prepare-a-production-environment/timeout-management.md` (4.12 and 4.13): adds a "Retry on a closed or reset connection" section describing the retry conditions (no body written yet, idempotent method), and updates the `keepAliveTimeout` and `http.requestTimeout`/`readTimeout` guidance to account for the retry doubling the worst-case `readTimeout` window.
- `analyze-and-monitor-apis/gateway-error-key-reference.md` (4.12 and 4.13): documents `GATEWAY_CLIENT_CONNECTION_RESET` alongside `GATEWAY_CLIENT_CONNECTION_CLOSED` as the same underlying pool-reuse race, and notes that from 4.12.18 both keys now only surface after the retry has already been attempted (or was ineligible).

## Testing

Rendered both changed pages locally to check the new section headings, internal anchor links, and the updated `{% hint %}` block.

[APIM-14873]: https://gravitee.atlassian.net/browse/APIM-14873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/apim/4.12/prepare-a-production-environment/timeout-management.md
```diff
@@ -71,6 +71,21 @@ When it closes a connection that carries a request, the failure surfaces as a `5
...
 {% endhint %}
 
+### Retry on a closed or reset connection
+
+**Available from 4.12.18.** A pooled connection can be closed or reset by the backend, or by an intermediary such as a load balancer, in the gap between the Gateway taking it from the pool and writing the next request to it — a race no timeout configuration eliminates, since a connection is only discovered to be dead by writing to it.
+
+The Gateway retries such a request once, on a fresh connection, when both of the following hold:
+
+* The Gateway hadn't yet written a body to the backend for this request.
+* The request method is idempotent: `GET`, `HEAD`, `OPTIONS`, `TRACE`, `PUT`, or `DELETE`.
```
phiz71: **Must fix.** The two conditions listed here aren't the two the connector applies, and the gap is not an edge case: `sendsRequestBody()` returns `true` whenever `request.version() == HTTP_2`, and `request` is `ctx.request()` — the consumer's request, not the one towards the backend. A plain `GET` received over h2 on the entrypoint is therefore never retried, whatever its method or body, and an HTTP/1.1 backend doesn't save it. The regression test `should_not_resend_an_http2_request_when_the_pooled_connection_was_closed_by_the_peer` pins exactly this.

A customer whose consumers speak h2 — a common case — upgrades to 4.12.18, sees no change, and this page tells them they should.

Same paragraph, smaller point: the body test is `hasBodyHeaders()`, i.e. the *presence* of `Content-Length` or `Transfer-Encoding`. A `GET` sent with `Content-Length: 0`, which some clients do, is excluded despite having no body. On a page whose job is to explain why a retry didn't happen, that deserves the half-sentence.

```suggestion
The Gateway retries such a request once, on a fresh connection, when all of the following hold:

* The Gateway hadn't yet written a body to the backend for this request. The test is the presence of a `Content-Length` or a `Transfer-Encoding` header, so a `GET` that carries `Content-Length: 0` is excluded even though it has no body.
* The request method is idempotent: `GET`, `HEAD`, `OPTIONS`, `TRACE`, `PUT`, or `DELETE`.
* The API consumer reached the Gateway over HTTP/1.1. An HTTP/2 request signals its body with data frames rather than with a header, so the Gateway treats every HTTP/2 request as carrying a body — a `GET` included — and never retries it. The version that decides is the one the consumer used, not the one configured on the endpoint.
```

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
```diff
@@ -71,6 +71,21 @@ When it closes a connection that carries a request, the failure surfaces as a `5
...
+
+A request that fails either condition — a `POST`, a `PATCH`, or any request sent with a body — isn't retried. Whether the request was ineligible for the retry or the retried attempt failed the same way, the failure is reported as `GATEWAY_CLIENT_CONNECTION_CLOSED` or `GATEWAY_CLIENT_CONNECTION_RESET`. See [Backend connection failures](../analyze-and-monitor-apis/gateway-error-key-reference.md#backend-connection-failures).
```
phiz71: **Must fix.** Whole classes of traffic get no retry at all, and the page scopes itself to none of them — its Overview says the opposite ("It applies to every API deployed on that Gateway").

The Limitations section of gravitee-io/gravitee-api-management#19143 already lists them:

* **v2-definition APIs / v4 emulation engine.** `InvokerAdapter` carries the same defect, but the connection handling lives in `io.gravitee.connector:gravitee-connector-http`, an external artifact the fix doesn't touch. Measured in that PR: 0.39–0.65% before, 0.44–0.50% after — unchanged.
* **gRPC.** `GrpcConnector` inherits the retry but always negotiates HTTP/2, so `sendsRequestBody` is always `true` and the path is inert.
* **WebSocket.** `WebSocketConnector` overrides `connect()` and doesn't inherit the retry.

Note the wording fix in the suggestion too: with a third condition above, "either condition" no longer holds.

```suggestion
A request that fails any of these conditions — a `POST`, a `PATCH`, or any request sent with a body — isn't retried. Whether the request was ineligible for the retry or the retried attempt failed the same way, the failure is reported as `GATEWAY_CLIENT_CONNECTION_CLOSED` or `GATEWAY_CLIENT_CONNECTION_RESET`. See [Backend connection failures](../analyze-and-monitor-apis/gateway-error-key-reference.md#backend-connection-failures).

The retry lives in the HTTP proxy endpoint of a v4 API. Three paths never retry, whatever the request looks like: an API with a v2 definition, which reaches its backend through a different connector; a gRPC endpoint, which always negotiates HTTP/2; and a WebSocket endpoint.
```

docs/apim/4.12/analyze-and-monitor-apis/gateway-error-key-reference.md
```diff
@@ -20,22 +20,24 @@ Two properties of these keys matter when you read them.
...
 {% endhint %}
 
-### On `GATEWAY_CLIENT_CONNECTION_CLOSED` and `GATEWAY_CLIENT_STREAM_ENDED_EARLY`
+### On `GATEWAY_CLIENT_CONNECTION_CLOSED`, `GATEWAY_CLIENT_CONNECTION_RESET`, and `GATEWAY_CLIENT_STREAM_ENDED_EARLY`
 
-These keys are misread more often than any other key. Neither means that the API consumer went away, and neither means that the Gateway timed out. Both mean the connection closed while the response was still incomplete in HTTP terms.
+These keys are misread more often than any other key. None of them means that the API consumer went away, and none means that the Gateway timed out. All three mean the connection closed or reset while the response was still incomplete in HTTP terms.
 
-The following two keys differ in **how far the response had progressed**:
+`GATEWAY_CLIENT_CONNECTION_CLOSED` and `GATEWAY_CLIENT_CONNECTION_RESET` are the same underlying event — the pooled connection was torn down by the peer before any response reached the API consumer — reported under different keys depending on whether that showed up as a clean close or a TCP reset. `GATEWAY_CLIENT_STREAM_ENDED_EARLY` differs in **how far the response had progressed**:
```
phiz71: **Must fix.** "The same underlying event" is sound as the rationale for retrying both keys — the load test in gravitee-io/gravitee-api-management#19143 makes that case well (a CLOSED-only retry left the reset failures untouched). But this is the reference page, where the sentence reads as what the key *means*, and there it's too broad: `ConnectionFailureClassifier` raises `CONNECTION_RESET` on a `StreamResetException` (HTTP/2 `RST_STREAM`) and on any "connection reset", on a brand-new connection as much as on a pooled one.

It also contradicts the table two lines above, which still reads "The backend reset the connection with a TCP RST, or sent an HTTP/2 `RST_STREAM`". Someone diagnosing a backend that rejects streams (`RST_STREAM`, `ENHANCE_YOUR_CALM`) is sent looking at connection pool reuse.

```suggestion
`GATEWAY_CLIENT_CONNECTION_CLOSED` and `GATEWAY_CLIENT_CONNECTION_RESET` are retried under the same conditions, because one teardown surfaces as a clean close or as a reset depending on the operating system and on timing. They still aren't interchangeable as a diagnosis: a reset also covers an HTTP/2 `RST_STREAM` from the backend, and a connection the backend resets on first use — neither of which involves connection reuse. `GATEWAY_CLIENT_STREAM_ENDED_EARLY` differs in **how far the response had progressed**:
```

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
```diff
@@ -71,6 +71,21 @@ When it closes a connection that carries a request, the failure surfaces as a `5
 `idleTimeout` is configured in milliseconds but applied in whole seconds. The remainder is discarded: `12200` ms becomes 12 seconds. Any value below `1000` ms becomes `0`, which **disables the timeout** instead of tightening it.
 {% endhint %}
 
+### Retry on a closed or reset connection
+
+**Available from 4.12.18.** A pooled connection can be closed or reset by the backend, or by an intermediary such as a load balancer, in the gap between the Gateway taking it from the pool and writing the next request to it — a race no timeout configuration eliminates, since a connection is only discovered to be dead by writing to it.
```
phiz71: **Should fix.** The predicate keys off the classified failure alone — never off where the connection came from. Any `CONNECTION_CLOSED`/`CONNECTION_RESET` raised before the response head resends, including on a freshly opened connection, and including a backend that received the request and then reset. Your own commit message says as much: "even with no body, the request head alone may have been received and acted on before the peer tore the connection down".

Two things follow that the page should own. A backend can legitimately see the same `GET` twice for a reason unrelated to pool reuse. And it is precisely that late-reset case — not the pool race — that consumes the `2 × readTimeout` this page then builds its new sizing rule on: a pool reuse race is detected on the write, in microseconds. As written, the two claims sit in tension.

```suggestion
**Available from 4.12.18.** A pooled connection can be closed or reset by the backend, or by an intermediary such as a load balancer, in the gap between the Gateway taking it from the pool and writing the next request to it — a race no timeout configuration eliminates, since a connection is only discovered to be dead by writing to it. The retry isn't restricted to that race: any connection closed or reset before the response headers arrive triggers it, on a connection just opened as much as on a reused one, and even when the backend had already received the request. That's why it's limited to requests that are safe to send twice.
```

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
```diff
@@ -95,9 +112,13 @@ This is the one inequality that governs correctness. Both timers run while the G
...
+{% endhint %}
+
 ### Keep `http.requestTimeout` above `readTimeout`
 
-The endpoint timeout produces the more precise diagnosis, so let it fire first. Set `http.requestTimeout` above `readTimeout` and below whatever the API consumers themselves tolerate.
+The endpoint timeout produces the more precise diagnosis, so let it fire first. A request eligible for [the retry on a closed or reset connection](timeout-management.md#retry-on-a-closed-or-reset-connection) can consume a fresh `readTimeout` window twice, so set `http.requestTimeout` above **2 × `readTimeout`** — and below whatever the API consumers themselves tolerate. Below that margin, a retried request can lose the race to `http.requestTimeout` and fail as `REQUEST_TIMEOUT` instead of getting the chance to complete.
```
phiz71: **Should fix.** The `2 ×` is stated as an unconditional rule, and from here it drives the recommended values further down — `Start from the defaults` (line 213), step 3 (line 219), the worked example `60000` → `70000` and its closing paragraph (line 255).

`http.requestTimeout` is Gateway-wide while `readTimeout` is per endpoint, so doubling the first relaxes the global backstop of every API to cover a case that is rare and, in the pool-reuse framing this page uses, near-instantaneous. Keep the rule — but say when the second window is actually consumed, so a reader sizes a margin instead of doubling by reflex.

```suggestion
The endpoint timeout produces the more precise diagnosis, so let it fire first. A request eligible for [the retry on a closed or reset connection](timeout-management.md#retry-on-a-closed-or-reset-connection) can consume a fresh `readTimeout` window twice, so `http.requestTimeout` has to clear **2 × `readTimeout`** for a retried request to reach that second window — below that margin it fails as `REQUEST_TIMEOUT` instead. That second window is only consumed when the connection breaks late, once the Gateway has already waited on the backend; the pool reuse race breaks it on the write. Weigh the margin against the fact that `http.requestTimeout` is Gateway-wide, while `readTimeout` is per endpoint: raising it relaxes the backstop of every API on that Gateway. And keep it below whatever the API consumers themselves tolerate.
```

docs/apim/4.12/analyze-and-monitor-apis/gateway-error-key-reference.md
```diff
@@ -20,22 +20,24 @@ Two properties of these keys matter when you read them.
...
-* `GATEWAY_CLIENT_CONNECTION_CLOSED`. Nothing had reached the API consumer yet, so the Gateway answers `502`.
+* `GATEWAY_CLIENT_CONNECTION_CLOSED` / `GATEWAY_CLIENT_CONNECTION_RESET`. Nothing had reached the API consumer yet, so the Gateway answers `502`.
+
+**From 4.12.18**, for a body-less idempotent request (`GET`, `HEAD`, `OPTIONS`, `TRACE`, `PUT`, or `DELETE` sent without a body), the Gateway already retried the request once, on a fresh connection, before reporting either of these two keys. Seeing `GATEWAY_CLIENT_CONNECTION_CLOSED` or `GATEWAY_CLIENT_CONNECTION_RESET` now means either that the request wasn't eligible for the retry — a `POST`, a `PATCH`, or any request sent with a body — or that the retried attempt failed the same way. Expect the volume of both keys to drop after upgrading: most pool-reuse races that used to surface as one of these keys now succeed silently on the retry instead. See [Retry on a closed or reset connection](../prepare-a-production-environment/timeout-management.md#retry-on-a-closed-or-reset-connection) for the retry's exact conditions.
```
phiz71: **Should fix.** "Most" extrapolates. The measurement behind it — 1.34–1.40% down to 0.005–0.05% over three runs a side — comes from `ab -k -c 50 -n 20000`, i.e. body-less HTTP/1.1 `GET`s against a v4 API: 100% eligible traffic. It's a solid number for the eligible slice and says nothing about a real mix, where `POST`, bodies, HTTP/2 and v2-definition APIs are all excluded.

Either drop the quantifier, or bound it to the eligible share.

```suggestion
**From 4.12.18**, for a body-less idempotent request (`GET`, `HEAD`, `OPTIONS`, `TRACE`, `PUT`, or `DELETE` sent without a body), the Gateway already retried the request once, on a fresh connection, before reporting either of these two keys. Seeing `GATEWAY_CLIENT_CONNECTION_CLOSED` or `GATEWAY_CLIENT_CONNECTION_RESET` now means either that the request wasn't eligible for the retry — a `POST`, a `PATCH`, or any request sent with a body — or that the retried attempt failed the same way. Expect the volume of both keys to drop after upgrading, in proportion to the share of traffic that qualifies for the retry. See [Retry on a closed or reset connection](../prepare-a-production-environment/timeout-management.md#retry-on-a-closed-or-reset-connection) for the retry's exact conditions.
```

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
```diff
@@ -80,7 +95,9 @@ Two properties are worth knowing:
...
-Backends that close idle connections without advertising the header are the case to configure for. Set `keepAliveTimeout` below the backend's own keep-alive window, so that the pool retires connections before the backend drops them. A connection retired by the backend while the Gateway still considers it usable produces a `GATEWAY_CLIENT_CONNECTION_CLOSED` on the next request that borrows it.
+Backends that close idle connections without advertising the header are the case to configure for. Set `keepAliveTimeout` below the backend's own keep-alive window, so that the pool retires connections before the backend drops them.
+
+A connection retired by the backend while the Gateway still considers it usable is, for a body-less idempotent request, now retried transparently on the next request that borrows it — see [Retry on a closed or reset connection](timeout-management.md#retry-on-a-closed-or-reset-connection). Tuning `keepAliveTimeout` is still worth doing: it avoids the wasted round trip the retry costs, and it remains the only protection for a `POST`, a `PATCH`, or any request sent with a body. It's no longer the only thing standing between this race and a client-visible `502`.
```
phiz71: Nit: the subject is the connection, but what gets retried is the request — and "on the next request that borrows it" is left over from the sentence this replaced, so the clause now attaches to the wrong noun.

```suggestion
The next request to borrow a connection the backend has already retired is now retried transparently, when it qualifies — see [Retry on a closed or reset connection](timeout-management.md#retry-on-a-closed-or-reset-connection). Tuning `keepAliveTimeout` is still worth doing: it avoids the wasted round trip the retry costs, and it remains the only protection for a `POST`, a `PATCH`, or any request sent with a body. It's no longer the only thing standing between this race and a client-visible `502`.
```

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
```diff
@@ -71,6 +71,21 @@ When it closes a connection that carries a request, the failure surfaces as a `5
 `idleTimeout` is configured in milliseconds but applied in whole seconds. The remainder is discarded: `12200` ms becomes 12 seconds. Any value below `1000` ms becomes `0`, which **disables the timeout** instead of tightening it.
 {% endhint %}
 
+### Retry on a closed or reset connection
```
phiz71, line 89: Placement: this sits under `## How each setting behaves`, between `Idle timeout` and `Keep-alive timeout` — but it isn't a setting and exposes no configuration, and it splits the one pair of timers the page keeps together everywhere else. Its own `##`, or the end of the section, would read better.

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
```diff
@@ -95,9 +112,13 @@ This is the one inequality that governs correctness. Both timers run while the G
 
 The default values satisfy this rule. Inverting them changes the diagnosis rather than the behavior. The request fails at the same point either way, but the reported cause then points at the backend.
 
+{% hint style="warning" %}
+For a body-less idempotent request, [the retry on a closed or reset connection](timeout-management.md#retry-on-a-closed-or-reset-connection) can't tell this self-inflicted closure apart from a genuine one by the backend — it only recognizes the Gateway's own explicit cancellations, such as a client abort, not a connection its own `idleTimeout` closed. The retried connection is governed by the same misconfiguration and closes the same way, so the closure is retried once: the caller waits roughly twice as long before the failure surfaces, and depending on `http.requestTimeout`, it can shift from a `502` (`GATEWAY_CLIENT_CONNECTION_CLOSED`) to a `504` (`REQUEST_TIMEOUT`).
+{% endhint %}
+
 ### Keep `http.requestTimeout` above `readTimeout`
```
phiz71: This heading now states the rule the section below revokes. Renaming it breaks two in-page anchors this very PR adds (line 85 and line 121), so it's a trade-off rather than a free fix — but leaving the old rule in the heading of the section that changes it is what readers will remember.

phiz71 (changes requested):
This documents a real behavior carefully, and most of what it says checks out against `HttpConnector` — the `WARN` line, the fresh `readTimeout` window per attempt, an `idleTimeout`-induced closure being retried, the anchors, the 4.12/4.13 parity, the arithmetic. Three things need fixing before it ships: two on which requests the retry actually covers, one on the error-key reference.

Comments are on the 4.12 files only; every one of them applies verbatim to the 4.13 copy, which is byte-identical.

One point that isn't a text defect, so it isn't inline: **4.12.18 isn't released** — `4.12.17` is the latest tag and `origin/4.12.x` carries `<revision>4.12.18</revision>` with the fix at HEAD. The 4.12 space is the live `documentation.gravitee.io/apim/`, so merging publishes behavior no shipped build has. "Available from 4.12.18" makes that explicit, so it's a coordination call — worth confirming before merge rather than an error.

Minor, for the record: the Jira key sits in the commit subject rather than as a link in the footer.

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
carlos-andres-osorio: Applied the suggested fix.

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
carlos-andres-osorio: Applied the suggested fix.

docs/apim/4.12/analyze-and-monitor-apis/gateway-error-key-reference.md
carlos-andres-osorio: Applied the suggested fix.

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
carlos-andres-osorio: Applied the suggested fix.

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
carlos-andres-osorio: Applied the suggested fix.

docs/apim/4.12/analyze-and-monitor-apis/gateway-error-key-reference.md
carlos-andres-osorio: Applied the suggested fix.

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
carlos-andres-osorio: Applied the suggested fix.

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
carlos-andres-osorio: Moved the Retry on a closed or reset connection section to the end of How each setting behaves, after Keep-alive timeout, restoring the Idle timeout / Keep-alive timeout adjacency. Applied to both the 4.12 and 4.13 copies.

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
carlos-andres-osorio: Renamed the heading to "Keep http.requestTimeout above twice readTimeout" and updated the one in-page cross-reference that pointed at the old anchor. Applied to both the 4.12 and 4.13 copies.

carlos-andres-osorio:
> top-level review: three must-fix items plus a coordination note that 4.12.18 is not yet released

Addressed all inline comments. Removed the "Available from 4.12.18" claim since 4.12.18 isn't tagged yet, rather than holding the PR.

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
```diff
@@ -80,10 +80,30 @@ Two properties are worth knowing:
...
 
 Like `idleTimeout`, this value is applied in whole seconds.
 
+### Retry on a closed or reset connection
+
+A pooled connection can be closed or reset by the backend, or by an intermediary such as a load balancer, in the gap between the Gateway taking it from the pool and writing the next request to it — a race no timeout configuration eliminates, since a connection is only discovered to be dead by writing to it. The retry isn't restricted to that race: any connection closed or reset before the response headers arrive triggers it, on a connection just opened as much as on a reused one, and even when the backend had already received the request. That's why it's limited to requests that are safe to send twice.
```
phiz71: **Must fix — this one is new.** Dropping the version marker isn't the fix for the release-timing note; it trades a scheduling question for a factual error. `4.12.18` now appears nowhere in the 4.12 docs, so both pages describe the retry unconditionally in the space that serves `documentation.gravitee.io/apim/` — i.e. to every reader from 4.12.0 on. Someone on 4.12.14 reads that the Gateway retries.

It also breaks the page's own convention: these two files still carry four `Available from 4.12.16` markers for `GATEWAY_CLIENT_STREAM_ENDED_EARLY`, one of them in the table eight lines above the paragraph you edited. One key is dated, the one next to it no longer is.

Keep the marker — it's what makes merging ahead of the tag harmless — and land the PR when `4.12.18` exists.

```suggestion
**Available from 4.12.18.** A pooled connection can be closed or reset by the backend, or by an intermediary such as a load balancer, in the gap between the Gateway taking it from the pool and writing the next request to it — a race no timeout configuration eliminates, since a connection is only discovered to be dead by writing to it. The retry isn't restricted to that race: any connection closed or reset before the response headers arrive triggers it, on a connection just opened as much as on a reused one, and even when the backend had already received the request. That's why it's limited to requests that are safe to send twice.
```

docs/apim/4.12/analyze-and-monitor-apis/gateway-error-key-reference.md
```diff
@@ -20,22 +20,24 @@ Two properties of these keys matter when you read them.
...
-* `GATEWAY_CLIENT_CONNECTION_CLOSED`. Nothing had reached the API consumer yet, so the Gateway answers `502`.
+* `GATEWAY_CLIENT_CONNECTION_CLOSED` / `GATEWAY_CLIENT_CONNECTION_RESET`. Nothing had reached the API consumer yet, so the Gateway answers `502`.
+
+For a body-less idempotent request (`GET`, `HEAD`, `OPTIONS`, `TRACE`, `PUT`, or `DELETE` sent without a body), the Gateway already retried the request once, on a fresh connection, before reporting either of these two keys. Seeing `GATEWAY_CLIENT_CONNECTION_CLOSED` or `GATEWAY_CLIENT_CONNECTION_RESET` now means either that the request wasn't eligible for the retry — a `POST`, a `PATCH`, or any request sent with a body — or that the retried attempt failed the same way. Expect the volume of both keys to drop after upgrading, in proportion to the share of traffic that qualifies for the retry. See [Retry on a closed or reset connection](../prepare-a-production-environment/timeout-management.md#retry-on-a-closed-or-reset-connection) for the retry's exact conditions.
```
phiz71: **Must fix.** The eligibility rewrite landed on `timeout-management.md` but not here, so the two pages now contradict each other. This one still opens on "For a body-less idempotent request (`GET`, `HEAD`, …), the Gateway already retried the request once", and still enumerates ineligibility as "a `POST`, a `PATCH`, or any request sent with a body".

For a consumer on HTTP/2, a v2-definition API, or a gRPC or WebSocket endpoint, that opening sentence is false. And this is the page someone lands on when the key shows up in their logs: the pointer to "the retry's exact conditions" at the end doesn't undo an enumeration stated inline — that's the one people read and act on.

The suggestion also restores the version marker, per the comment on `timeout-management.md`.

```suggestion
**From 4.12.18**, a request that qualifies for the retry — body-less, on an idempotent method, received over HTTP/1.1, against the HTTP proxy endpoint of a v4 API — was already retried once, on a fresh connection, before either of these two keys was reported. Seeing `GATEWAY_CLIENT_CONNECTION_CLOSED` or `GATEWAY_CLIENT_CONNECTION_RESET` therefore means either that the request wasn't eligible — a `POST` or a `PATCH`, a request sent with a body, a consumer on HTTP/2, a v2-definition API, a gRPC or WebSocket endpoint — or that the retried attempt failed the same way. Expect the volume of both keys to drop after upgrading, in proportion to the share of traffic that qualifies for the retry. See [Retry on a closed or reset connection](../prepare-a-production-environment/timeout-management.md#retry-on-a-closed-or-reset-connection) for the retry's exact conditions.
```

phiz71 (changes requested):
Thanks — the nine comments from the first round are all handled, and I checked the two that could go wrong: the heading rename leaves no dangling reference to the old anchor anywhere in `docs/`, and the 4.12/4.13 copies are still byte-identical.

Two things left, both one paragraph each. One is a regression from this update: removing `Available from 4.12.18` answers the scheduling question by making the pages inaccurate for every 4.12 patch before it. The other is the eligibility rewrite landing on one of the two pages only, so `gateway-error-key-reference.md` now contradicts `timeout-management.md`.

Comments on the 4.12 files again; both apply verbatim to the 4.13 copy.

docs/apim/4.12/analyze-and-monitor-apis/gateway-error-key-reference.md
carlos-andres-osorio: applied

docs/apim/4.12/prepare-a-production-environment/timeout-management.md
carlos-andres-osorio: applied

carlos-andres-osorio: @gravitee-io/tech-writers FYI.

As far as I know, the latest APIM release was 4.12.17 (https://graviteeio.slack.com/archives/C02NGT20S4W/p1787323398239029), so this change will be part of 4.12.18.

If that's not the case, there are a few places in this PR that will have to be updated to match the release that contains these changes.

Please let me know if changes are needed.

docs/apim/4.13/prepare-a-production-environment/timeout-management.md
```diff
@@ -80,10 +80,30 @@ Two properties are worth knowing:
...
 
 Like `idleTimeout`, this value is applied in whole seconds.
 
+### Retry on a closed or reset connection
+
+A pooled connection can be closed or reset by the backend, or by an intermediary such as a load balancer, in the gap between the Gateway taking it from the pool and writing the next request to it — a race no timeout configuration eliminates, since a connection is only discovered to be dead by writing to it. The retry isn't restricted to that race: any connection closed or reset before the response headers arrive triggers it, on a connection just opened as much as on a reused one, and even when the backend had already received the request. That's why it's limited to requests that are safe to send twice.
```
phiz71: The `Available from 4.12.18.` marker went back onto the 4.12 copy but not this one, so the 4.13 space now contradicts itself: its `gateway-error-key-reference.md` says **From 4.12.18**, this page says nothing. My fault for commenting on the 4.12 files only — the suggestion button applies to one file, and this is the one it couldn't reach.

```suggestion
**Available from 4.12.18.** A pooled connection can be closed or reset by the backend, or by an intermediary such as a load balancer, in the gap between the Gateway taking it from the pool and writing the next request to it — a race no timeout configuration eliminates, since a connection is only discovered to be dead by writing to it. The retry isn't restricted to that race: any connection closed or reset before the response headers arrive triggers it, on a connection just opened as much as on a reused one, and even when the backend had already received the request. That's why it's limited to requests that are safe to send twice.
```

phiz71 (commented):
One line left over from applying the suggestions: the version marker is missing from the 4.13 copy of `timeout-management.md`. Everything else from both rounds checks out — I re-read the final text of all four files.

docs/apim/4.13/prepare-a-production-environment/timeout-management.md
carlos-andres-osorio: applied

phiz71 (approved):
Approving the content — everything from the three rounds is addressed, and I re-read the final text of all four files rather than the diff.

Verified on `b12ac75`: the 4.12 and 4.13 copies are byte-identical again on both pages, the `4.12.18` marker is on all four files, the renamed heading's anchor follows the rule this page already demonstrates twice (`### HTTP/2` → `#http-2`), no reference to the old anchor survives anywhere under `docs/`, and the seven checks are green.

> [!WARNING]
> **Don't merge until 4.12.18 is released.** The latest tag is `4.12.17`, and `origin/4.12.x` still carries `<revision>4.12.18</revision>` with the retry at HEAD — the behavior these pages describe exists in no shipped build yet. The `./docs/apim/4.12` space is the live `documentation.gravitee.io/apim/`, so merging publishes it immediately. The `Available from 4.12.18` marker keeps that from being wrong, but it still puts a page in front of customers describing something they can't get. Wait for the tag, then merge.

One thing I deliberately left alone, in case you'd rather close it while you're here: the warning callout under `Keep readTimeout shorter than idleTimeout` still says "For a body-less idempotent request", the two-condition shorthand from before the rewrite. The section above it now defines three, so "for a request that qualifies for the retry" would match. Not worth another round on its own.

garethbrinn (approved):

PR #2261, opened 2026-08-25
docs(gamma): document environment-scoped metadata in Platform Managem…

tolaore:
https://gravitee.atlassian.net/browse/FOUND-3


docs/gamma/4.13/platform-management/manage-environment-metadata.md
```diff
@@ -0,0 +1,101 @@
...
+## Add a metadata entry
+
+To add an entry, complete the following steps:
+
+1. Select **Add Global Metadata**.
+2. Enter the entry details described in the following table:
+
+    | Field      | Description                                                                                                    | Required |
+    | ---------- | -------------------------------------------------------------------------------------------------------------- | -------- |
+    | **Name**   | The entry's name. Gravitee generates the key from it, so a name that produces a key an entry in the environment already uses is rejected. | Yes      |
```
arpzilla-dev: This overstates create-time uniqueness.

The Gamma create form does not check for an existing name or key (`MetadataSheet.tsx` / `useCreateMetadata`). On the backend, `MetadataServiceImpl.create` validates format and value only — there is no `DuplicateMetadataNameException` on create.

Duplicate **name** is rejected only on **update** (`MetadataServiceImpl.update`, case-insensitive name, different key).

A create that generates a key that already exists would fail at the repository/DB layer, not with the friendly “already uses” rejection described here. Two different names can also generate the same key via `IdGenerator.generate(name)`, and that path is also unhandled in `create`.

Please either:
1. Limit this sentence to rename/update (which the Edit section already covers correctly), or
2. Confirm a UI/API error you actually saw on create and document that error instead.

docs/gamma/4.13/platform-management/manage-environment-metadata.md
```diff
@@ -0,0 +1,101 @@
...
+
+| Format      | Value                                                                                                                    |
+| ----------- | -------------------------------------------------------------------------------------------------------------------------- |
+| **String**  | Any text.                                                                                                                  |
+| **Numeric** | A number, entered in a number field.                                                                                       |
+| **Boolean** | `true` or `false`, selected from a dropdown. Selecting the format sets the value to `false`.                                |
+| **Date**    | A date, entered in a date field. The platform stores the date part alone, in `yyyy-MM-dd` form, and rejects a date that doesn't exist. |
+| **Mail**    | An email address. The console reads **Invalid email** under the field while the value isn't one.                            |
+| **URL**     | A URL. Enter it in full, including `http://` or `https://`. The console reads **Invalid URL** under the field while the value isn't one. |
```
arpzilla-dev: The “enter it in full, including `http://` or `https://`” rule is true for the **backend**, not clearly for the **console regex**.

Backend `checkMetadataFormat` for `URL` does `new URL(decodedValue)`, which requires a scheme.

Gamma UI uses:
```ts
const URL_REGEX = /^((\$\{.+\})|(https?:\/\/)?([\da-z.-]+)\.([a-z.]{2,6})\b(...))$/;
```
The `https?:\/\/` part is **optional**, so `example.com` can pass `isValueValid` / “Invalid URL”.

The value input is also `type="url"`. Native HTML5 URL checks often require a scheme, so the live field may still block schemeless values — I have not confirmed that in the browser.

Please try `example.com` vs `https://example.com` in the create sheet and document the actual console rule. If the regex is the source of “Invalid URL”, the “must include http(s)” sentence is too strong.

docs/gamma/4.13/platform-management/manage-environment-metadata.md
```diff
@@ -0,0 +1,101 @@
...
+
+To add an entry, complete the following steps:
+
+1. Select **Add Global Metadata**.
+2. Enter the entry details described in the following table:
+
+    | Field      | Description                                                                                                    | Required |
+    | ---------- | -------------------------------------------------------------------------------------------------------------- | -------- |
+    | **Name**   | The entry's name. Gravitee generates the key from it, so a name that produces a key an entry in the environment already uses is rejected. | Yes      |
+    | **Format** | **String**, **Numeric**, **Boolean**, **Date**, **Mail**, or **URL**. **String** is preselected, and the format is fixed after creation. | Yes      |
```
arpzilla-dev: Accurate for the Gamma console: the format `Select` is `disabled={mode === 'edit'}`, and the helper text is “Format cannot be changed after creation.”

It is **not** a backend invariant. `MetadataServiceImpl.update` still takes `format` from `UpdateMetadataEntity` and persists it. A direct Management API client could change format.

Fine to leave as written if this page is console-only. If you want it to stay true for API users too, say “In the console, format cannot be changed after creation.”

docs/gamma/4.13/platform-management/manage-environment-metadata.md
```diff
@@ -0,0 +1,101 @@
...
+* **Key**. The identifier that APIs and Developer Portal pages use to reference the entry. Gravitee generates it from the name when the entry is created.
+* **Name**. The label the console displays.
+* **Format**. The kind of value the entry holds: String, Numeric, Boolean, Date, Mail, or URL.
+* **Value**. The value APIs inherit.
+
+The Metadata page in the Gamma console lets you view, add, edit, and delete the entries of the selected environment.
+
+## View environment metadata
+
+From the Gamma console sidebar, select **Platform Management**, open the **Environment** section, and then navigate to **Metadata**.
```
arpzilla-dev: The path works, but it skips a group. In Gamma, Metadata sits under **Platform Management → Environment → APIs & Assets → Metadata** (`gravitee-gamma-module-platform/.../config/navigation.ts`).

Adding **APIs & Assets** would match the sidebar. Not a blocker.

docs/gamma/4.13/platform-management/manage-environment-metadata.md
tolaore: Confirmed, and fixed in 269114a7. I took your option 1 for the field table and then documented what the create path actually does, verified on a local master-line stack (platform module 4.13.0-SNAPSHOT).

- The **Name** row now says only that Gravitee generates the entry's key from the name. Duplicate-name rejection stays in the Edit section, where `MetadataServiceImpl.update` does check it (HTTP 400, `metadata.exists`, "The metadata '<name>' already exists." — confirmed live).
- What create does, confirmed in the console and over the API: no client-side check, the request goes out, and a name whose generated key already exists comes back HTTP 500 with the sheet still open, no entry created, and the toast reading "An error occurred while trying to create metadata <name>" (`notify.error` falls through to the API message via `extractErrorMessage`).

The page now reads: "The console doesn't flag a name that's already in use before you select **Add**. When the name generates a key that another entry in the environment already uses, the entry isn't created. The console then reads **An error occurred while trying to create metadata**, followed by the name."

The unguarded create path is on the findings list for the SME as a probable bug (a 500 where the update path returns a friendly 400), separately from this PR.

docs/gamma/4.13/platform-management/manage-environment-metadata.md
tolaore: You were right that the sentence attributed the rule to the wrong layer. I ran the comparison you asked for, in the create sheet on a local master-line stack (platform module 4.13.0-SNAPSHOT), driving the real console:

- `example.com`, format URL: no **Invalid URL** message, **Add** enabled (the regex scheme group is optional, as you noted). Clicking **Add** sends no request at all: the value input is `type="url"`, so the browser's constraint validation blocks the submit (`validity.typeMismatch: true`, `validationMessage: "Please enter a URL."` in Chromium). Sheet stays open, no toast, no POST.
- `https://a.example.com`: valid, POST goes out, entry created.
- API path independently: `POST .../configuration/metadata` with `example.com` returns HTTP 500, "Error occurred while trying to validate format URL of value example.com" (`new URL(decodedValue)` in `checkMetadataFormat`).

So the console rule is: **Invalid URL** covers a value that isn't URL-shaped, and a missing scheme is caught by the field itself rather than by that message. The row and the note now say exactly that:

| **URL** | A URL, entered in full, including `http://` or `https://`. The console reads **Invalid URL** under the field while the value isn't URL-shaped. |

"The **Invalid URL** message covers a value that isn't URL-shaped. A value that omits the scheme, such as `example.com`, clears that check and leaves **Add** enabled, and the browser blocks the submission instead. Gravitee rejects a URL value that has no scheme as well, so enter the URL in full."

docs/gamma/4.13/platform-management/manage-environment-metadata.md
tolaore: Scoped to the console, thanks. I confirmed the API side rather than taking it on the code read alone: `PUT .../configuration/metadata` for an existing key with a different `format` returns HTTP 200 and persists the new format, so the invariant really is UI-only (`convert(ExecutionContext, UpdateMetadataEntity)` copies the format straight from the entity).

The **Format** row now reads: "**String**, **Numeric**, **Boolean**, **Date**, **Mail**, or **URL**. **String** is preselected. After the entry is created, the console shows **Format** as read-only."

That leaves the claim tied to the surface the page documents, and it matches the Edit section, which already says **Key** and **Format** are read-only.

docs/gamma/4.13/platform-management/manage-environment-metadata.md
tolaore: Added. The navigation step now reads: "From the Gamma console sidebar, select **Platform Management**, and open the **Environment** section. Under **APIs & Assets**, select **Metadata**."

Split into two sentences rather than a breadcrumb chain, per the house rule against arrow navigation. Verified against `navigation.ts` (Environment section, **APIs & Assets** group, `metadata` item with its label from `routes.ts`) and against the capture already in the PR, where **APIs & Assets** renders as the group heading above **Metadata**.

One follow-up worth noting: the sibling pages don't name their group either (`monitor-gateway-instances.md` says "open the **Environment** section, and then navigate to **Gateways**", where Gateways sits under **System & Security**). I've left those alone in this PR and raised it as a separate consistency pass.

arpzilla-dev (approved):

garethbrinn (approved):

PR #2267, opened 2026-08-26
docs(gamma): document the MCP Proxy workspace tabs for 4.12

garethbrinn:
Four tabs of the **MCP Proxy** workspace ship in 4.12 but are documented only in 4.13. Since 4.12 is the published default, readers on the default version have no pages for screens that exist. This ports them back.

| Page | Console tab |
|---|---|
| Configure MCP Proxy notifications | `GENERAL → Notifications` |
| Configure MCP Proxy deployment | `OPERATIONS → Deployment → Configuration` |
| Review MCP Proxy deployment history | `OPERATIONS → Deployment → History` |
| Configure logging and tracing | `GATEWAY → Reporter Settings` |

## Verified, not carried over on trust

Every page was checked against a running **4.12.16** console:

- The workspace rail matches: General with Overview, Settings, API Properties, Notifications and Endpoint; Gateway with Reporter Settings; Design with Policy Studio; Consumer access with Plans and Consumers; Monitoring with Audit Logs; Observability; Operations.
- The notifications table and its two default rows, the sharding-tag copy, the deployment history columns, and the reporter settings groups match the pages word for word.
- **The one claim needing interaction was tested directly:** the Span Attribute Redaction section appears only when Trace enabled *and* Verbose are both on. Confirmed, then discarded without saving.

The four screenshots were reviewed before inclusion and match this build.

## Not verified

Three details are inherited from the 4.13 text and were not independently confirmed: the notification permission codes, the add-notification flow beyond the button, and the Grafana reference in the tracing page, which brushes against the rule on documenting third-party products.

The A2A Proxy equivalents are in a separate PR, and a Platform Management page in a third.


docs/gamma/4.12/agent-management/build/configure-mcp-proxy-notifications.md
```diff
@@ -0,0 +1,53 @@
...
+
+To add a notification, follow these steps:
+
+1. Click **Add notification**.
+2. Enter a **Name**.
+3. Select a **Channel**. The list offers the notifiers configured for your organization, such as **Default Email Notifier**. The channel sets the target field: **Email list** for an email notifier, or **Webhook** for a webhook notifier.
+4. Enter the target. Separate multiple email addresses with a space, a comma, or a semicolon. The field supports Expression Language.
+5. For a webhook, optionally enable **Use system proxy** to route webhook calls through the gateway's configured system proxy.
+6. Under **Events**, select the API events that send a notification. Events are grouped by category, such as **API KEY** and **SUBSCRIPTION**.
+7. Optional: under **Groups**, select the groups the notification applies to. Leave it empty to apply the notification to all groups.
```
wbabyte, line 37: Groups isn't part of the add flow. In `DeployableNotificationFormPage.tsx` the Groups block is gated on `form.isPortalEdit && form.groups.length > 0`, so it renders only when editing the **Portal** notification — never when adding one. Suggest dropping this step and covering Groups in the edit section instead.

docs/gamma/4.12/agent-management/build/configure-mcp-proxy-notifications.md
```diff
@@ -0,0 +1,53 @@
...
+8. Click **Add notification**.
+
+## Edit or delete a notification
+
+To change a notification, open its **Actions** menu, and then click **Edit**. You can adjust its target, its selected events, and its groups, and then click **Save**. The channel of an existing notification can't be changed. The form shows it as a read-only **Email** or **Webhook** badge.
```
wbabyte, line 42: "adjust its target, its selected events, and its groups" — groups is only editable on the **Portal** notification (same `isPortalEdit` gate), and even there the checkboxes are disabled unless you're the primary owner ("Only the primary owner can change groups."). A generic email/webhook notification has no groups control at all. Suggest splitting: target + events for any notification, groups only for the Portal notification and only as primary owner.

docs/gamma/4.12/agent-management/build/configure-mcp-proxy-deployment.md
```diff
@@ -0,0 +1,49 @@
...
+4.  Click **Configuration**.
+
+    <figure><img src="../../.gitbook/assets/gamma-mcp-proxy-deployment-configuration.png" alt="The Deployment Configuration page for MCP Proxy, showing the Sharding tags card with one tag selected as a chip and the tag list open"><figcaption><p>The Deployment Configuration page</p></figcaption></figure>
+
+## Assign sharding tags
+
+The **Sharding tags** card lists the tags defined for your organization. To assign them, follow these steps:
+
+1. Open the tag list and select a tag. Each row shows the tag name and, when the tag has one, its description.
+2. Repeat for each tag you want to assign. Every selected tag appears as a chip beneath the list.
```
wbabyte, line 25: In 4.12 the chips render *inside* the combobox input, not beneath the list — the `1.x` component uses `ComboboxChips` wrapping `ComboboxChipsInput placeholder="Select tags…"`, with the dropdown opening below it. The screenshot on line 18 shows this: the `external ×` chip sits in the input field with the tag list open underneath. Suggest "Every selected tag appears as a chip in the field." (For context, the chips did move to a separate "Configured Sharding Tags" box on `main`, so this wording may be correct for a later version — but not for 4.12.)

docs/gamma/4.12/agent-management/build/configure-mcp-proxy-deployment.md
```diff
@@ -0,0 +1,49 @@
...
+3. Optional: to unassign a tag, click the remove control on its chip.
+4.  Click **Save changes**, or **Discard** to revert.
+
+The **Discard** and **Save changes** buttons appear only once your selection differs from the saved value. A successful save confirms with **Deployment configuration saved**.
+
+When your organization has no tags, the card reads **No sharding tags configured**. Sharding tags are managed at the organization level, on the **Entrypoints & Sharding Tags** page in Platform Management.
```
wbabyte, line 31: Two things here. The 4.12 product string in this exact empty state reads "Sharding tags are managed at the organisation level under Gateway settings", and the sibling 4.12 page `api-management/build/configure-your-api-proxy/configure-deployment.md:29` says "under the Gateway settings" — so 4.12 now points readers at two different places. The Platform Management wording is probably the more accurate one, but the two pages should agree. Second, the **Entrypoints & Sharding Tags** page doesn't exist in 4.12 until the third PR lands, so until then a reader has nothing to navigate to.

wbabyte (approved):
Validated the port against the module source at the version 4.12 actually ships, plus the APIM backend. Approving — four fixes inline, none blocking the port itself.

**Version check holds up.** `gravitee-api-management@4.12.x` pins `gravitee-gamma-module-aim` **1.13.0**, and on the module's `1.x` line the Notifications table still has Name / Notifier / Subscribed events / Groups. The rename to Channel/Events/Target landed on `main` in `134d74b4` (AIAM-127, "Dropped the Groups column") and was never backported to `1.x`. So the text is right for 4.12, and the four screenshots match `1.x` exactly — rail, table headers, `All groups`, Items-per-page, and the Reporter Settings groups.

Heads-up beyond this PR: the 4.13 pages are byte-identical, so if 4.13 ships module >= 3.15.1 they are now stale there (columns renamed, Groups dropped, plus summary stat cards, a "Configured notifications" card wrapper, and a "Why configure notifications?" card the text doesn't cover).

**Mechanics are clean.** All four `.md` files and all four PNGs are byte-identical to the 4.13 originals (same blob SHAs). Image relative paths resolve, `../../observe/monitor-your-mcp-servers.md` exists in 4.12, frontmatter matches sibling pages, and the SUMMARY nesting is correct — tidier than 4.13, whose Atlassian/Stripe entries got orphaned under the deployment pages.

**On the "Not verified" list:** two of the three are verifiable and check out. The permission codes are exactly right (`api-notification-c` / `-u` / `-d` in `notificationPermissions.ts`), and the add-notification flow is right apart from the Groups step flagged below. The Grafana mention is the product's own string ("Enables log-to-trace linking in Grafana and other OTel-compatible backends"), so the page is quoting the UI.

**Also verified:** the rollback warning hint is correct and well earned — `RollbackApiUseCase` on `4.12.x` updates the definition plus plans and writes an audit log, never a publish event. Worth noting that the most visible misleading string isn't the dialog but the success toast, which reads "Rolled back and redeployed".

**Scope gap for a follow-up:** LLM Proxy has the same four 4.13-only pages missing from 4.12 (`configure-llm-proxy-notifications`, `configure-llm-proxy-deployment`, `configure-llm-proxy-logging-and-tracing`, `review-llm-proxy-deployment-history`), driven by the same shared components. The description promises A2A and Platform Management follow-ups but not LLM.

PR #2269, opened 2026-08-26
docs(gamma): document environment metadata in Platform Management for 4.12

garethbrinn:
The **Metadata** screen ships in the 4.12 console and holds live data, but only 4.13 documents it. 4.12 is the published default, so readers on the default version have no page for it.

## Verified against a running 4.12.16 console

The screen sits under `APIs & Assets → Metadata`, with the four columns the page describes — Key, Name, Format, Value — an Actions menu, the **Add Global Metadata** action, and the shipped default entry `email-support` of format Mail.

## The 4.13 screenshot is deliberately omitted

That image shows a Platform Management sidebar containing **Shared Policy Groups, Gateways, Alerts, Security Plan Types and Audit**. None of those screens exists in the 4.12 build, whose sidebar holds only Applications, Metadata, Dictionaries and Access Management. Including it would show readers navigation they do not have. The prose stands on its own.

**Worth checking separately:** whether that screenshot is accurate for 4.13 either, or whether it was captured on a build ahead of both.

## One claim not verified

The page states that a Developer Portal support request is refused while the `email-support` entry still holds its shipped default. That behaviour is inherited from the 4.13 text and was not exercised here. The default value itself is the product's own placeholder and was confirmed on the screen.


docs/gamma/4.12/platform-management/manage-environment-metadata.md
```diff
@@ -0,0 +1,102 @@
...
+* **Key**. The identifier that APIs and Developer Portal pages use to reference the entry. Gravitee generates it from the name when the entry is created.
+* **Name**. The label the console displays.
+* **Format**. The kind of value the entry holds: String, Numeric, Boolean, Date, Mail, or URL.
+* **Value**. The value APIs inherit.
+
+The Metadata page in the Gamma console lets you view, add, edit, and delete the entries of the selected environment.
+
+## View environment metadata
+
+From the Gamma console sidebar, select **Platform Management**, and open the **Environment** section. Under **APIs & Assets**, select **Metadata**.
```
arpzilla-dev: Suggestion: readers of this page won’t have an Environment section — that sidebar grouping is intended for 4.13. Metadata sits under APIs & Assets next to Dictionaries (with Applications under Overview, and Access Management under System & Security).

Could we phrase this the same way as the dictionaries/applications pages, for example:

From the Gamma console sidebar, select **Platform Management**. Under **APIs & Assets**, select **Metadata**.

Omitting the 4.13 screenshot already avoids showing that later nav; this line is the one leftover that still points at Environment.

ps: Good for this branch, but keep this review for 4.13

arpzilla-dev (approved):

PR #2282, opened 2026-08-31
docs(apim): document the kafka_active_connections gauge

ThibaudAV:
https://gravitee.atlassian.net/browse/ESM-2

`kafka_active_connections` shipped in 4.13 with nothing in the documentation. It is **opt-in and off by default**, so an operator had no way to discover it — the only description lived in the plugin README, which nobody reads to configure a Gateway.

Adds a section to `4.13/analyze-and-monitor-apis/logging/expose-metrics-to-prometheus.md`, between "Configure labels" and "Prometheus configuration".

## What it covers

- what the gauge measures, and how it differs from the connection analytics — *what is connected now* against *what happened*;
- how to enable it, in the same three deployment shapes the page already uses (`gravitee.yml`, `.env`, Helm);
- a sample scrape line with the real labels;
- **cardinality** — series are removed when their last connection closes, which is what makes `client_id` usable as a label at all, plus what `maxClientIdCardinality` actually caps and what `client_id="other"` means;
- **virtual clusters** — the bootstrap connection is not counted, so the gauge undercounts by one per client on those topologies.

## On the sample line

Transcribed from `KafkaActiveConnectionsGauge` rather than paraphrased. It is `api_id`, not `api`; and `application` is the reporting component while `application_id` is the subscribing application — easy to conflate, and this is the line readers paste into Grafana.

Also spells out that `client_software_name="unknown"` is a *value*, not an absence: it is what a client advertising no library (KafkaJS, or anything on `ApiVersions` older than v3) reports, and those connections are counted like any other.

## Not here

The 4.12 page. The gauge ships in the `7.1.x` reactor, which is 4.13 — a 4.12 section will follow the backport if that lands.


docs/apim/4.13/analyze-and-monitor-apis/logging/expose-metrics-to-prometheus.md
```diff
@@ -84,6 +84,86 @@ services:
...
+          enabled: true
+          maxClientIdCardinality: 10000
+```
+{% endtab %}
+{% endtabs %}
+
+A scrape then returns one series per client identity:
+
+```
+kafka_active_connections{application="gateway",organization_id="DEFAULT",environment_id="DEFAULT",api_id="orders-stream",plan_id="p1",application_id="a1",client_id="orders-consumer-1",client_software_name="librdkafka",client_software_version="2.6.1"} 42
```
callaertanthony: `application="gateway"` is not the value that gets emitted.

The gauge builds the tag from `node.application()`, and the Gateway's `GatewayNode` returns `APPLICATION_NAME = "gio-apim-gateway"`. It is the same value `VertxFactory` sets as a common tag on every other Gateway metric — which is exactly the claim the next paragraph makes, so the paragraph is right and only the sample disagrees with it.

This is the line a reader pastes into Grafana, and a matcher on `application="gateway"` returns an empty result.

```suggestion
kafka_active_connections{application="gio-apim-gateway",organization_id="DEFAULT",environment_id="DEFAULT",api_id="orders-stream",plan_id="p1",application_id="a1",client_id="orders-consumer-1",client_software_name="librdkafka",client_software_version="2.6.1"} 42
```

docs/apim/4.13/analyze-and-monitor-apis/logging/expose-metrics-to-prometheus.md
```diff
@@ -84,6 +84,86 @@ services:
...
+
+A client that does not advertise its library — KafkaJS, or any client on an `ApiVersions` request older than v3 — is reported as `client_software_name="unknown"`, and likewise for the version. That is a value, not an absence: those connections are counted like any other.
+
+### Cardinality
+
+A series is **removed from the registry when its last connection closes**. The number of live series therefore follows how many connections are open at once, not how many client ids have ever been seen — which is what makes `client_id` usable as a label at all, since it is free-form text the client chooses and often carries a pod name or a UUID. A client that reconnects under a new id leaves nothing behind.
+
+`maxClientIdCardinality` (default `10000`) caps how many distinct `client.id` values are labelled at the same time, per API. Beyond it, further client ids are reported as `client_id="other"`. The connection is still counted, so the gauge keeps summing to the true number of active connections and only loses granularity. Slots are released as connections close, so a Gateway returns to full granularity on its own — the cap is a backstop, not a budget that runs out.
```
callaertanthony, line 151: Worth saying here that the cap covers `client_id` and nothing else.

`KafkaActiveConnectionsGauge` calls this out explicitly: `client_software_name` and `client_software_version` are bounded in charset and length but not in distinct-value count. They come off the unauthenticated `ApiVersions` frame, so a client that puts a build hash or an instance id in its version string produces one series per value, with no bucket to collapse onto — `maxClientIdCardinality` does not touch them.

As written, the section reads as though the ceiling bounds the whole metric. On a page whose existing hint warns about cardinality blowups, that is the wrong impression to leave. Removal-on-last-close still keeps it proportional to concurrent connections rather than values ever seen, so it is a caveat rather than a warning — two sentences after this paragraph would cover it.

docs/apim/4.13/analyze-and-monitor-apis/logging/expose-metrics-to-prometheus.md
```diff
@@ -84,6 +84,86 @@ services:
...
+
+A `client.id` longer than 256 characters is truncated and suffixed with `...`, which keeps it correlatable with the client's own configuration and distinguishable from `client_id="other"`, which means the cap was reached instead.
```
callaertanthony: Minor: the suffix is inside the ceiling, not added to it.

`KafkaClientId.MAX_LENGTH` is 256 and its javadoc describes it as "length ceiling, suffix included" — so a long id comes back as 253 characters plus `...`, 256 in total. "truncated and suffixed with `...`" reads as 259.

Suggest: *A `client.id` longer than 256 characters is cut to 256, the last three being `...`, which keeps it correlatable […]*

callaertanthony (commented):
Three findings. The first one blocks — the rest of the section checks out line by line against `KafkaActiveConnectionsGauge`, including tag order, the `application` vs `application_id` distinction, the `max + 1` ceiling, the overflow-still-counted behaviour and the virtual-cluster undercount.

Also verified: the Helm block does reach `gravitee.yml` (`gateway-configmap.yaml` renders `gateway.services.metrics` with a generic `toYaml`, so the nested `kafka.activeConnections` passes through); the `.env` camelCase spelling matches the convention used across the 4.13 docs; and the `.version-overrides` entry is justified — 4.12.x pins reactor 7.0.8, master pins 7.1.0-alpha.13, which is the first tag carrying the gauge.

ThibaudAV:
All three taken. The first one is the finding I am most annoyed to have needed.

## `application="gateway"` — wrong, and wrong twice over

Verified: `GatewayNode.APPLICATION_NAME = "gio-apim-gateway"`, and the gauge builds the tag from `node.application()`. A Grafana matcher on `application="gateway"` returns nothing.

What makes this worth naming rather than just fixing: this is the **second** wrong label in that same line. An earlier pass corrected `api` to `api_id` by transcribing from `KafkaActiveConnectionsGauge` — and I still shipped this one, because the gauge shows `node.application()`, not its value, so I filled in the value from expectation instead of following the indirection. Transcribing what is on screen is not enough when what is on screen is a reference.

And you are right that the paragraph underneath already asserted the correct thing — the sample was the only part disagreeing with it.

## The cap's scope — the most useful of the three for a reader

Confirmed: `reserveClientId` claims a slot for `client_id` and nothing else. `client_software_name` and `client_software_version` are bounded in charset and length but not in distinct-value count, and they come off the same unauthenticated frame, so a client putting a build hash in its version string yields one series per value with no bucket to collapse onto.

My section implied the ceiling bounded the whole metric — on a page whose existing hint warns about cardinality blowups, exactly the wrong impression. Added, framed as a caveat rather than a warning, and closed on removal-on-last-close keeping the total proportional to concurrent connections, as you suggested.

## The suffix is inside the ceiling

Right: `cut = MAX_LENGTH - TRUNCATION_SUFFIX.length()`, so 253 characters plus `...` is 256 in total. "truncated and suffixed" reads as 259. Reworded to your phrasing.

---

Thanks for checking the Helm rendering path, the `.env` casing and the `.version-overrides` justification against the pinned reactor versions — those were the parts I could not verify from the docs repo alone.

`lint_overrides.py . apim 4.12 4.13` still passes. Squashed into the single commit.

docs/apim/4.13/analyze-and-monitor-apis/logging/expose-metrics-to-prometheus.md
```diff
@@ -84,6 +84,88 @@ services:
...
+
+The cap applies to `client_id` alone. `client_software_name` and `client_software_version` are bounded in charset and length but not in how many distinct values they may take: they come off the same unauthenticated handshake frame, so a client that puts a build hash or an instance id in its version string produces one series per value, with no bucket to collapse onto. Removal on last close still keeps the total proportional to the connections open at once rather than to every value ever seen.
+
+A `client.id` longer than 256 characters is cut to 256, the last three being `...`. That keeps it correlatable with the client's own configuration and distinguishable from `client_id="other"`, which means the cap was reached instead.
+
+{% hint style="info" %}
```
callaertanthony, line 157: Non-blocking style nit, take it or leave it.

This hint opens with "The cap applies to the `client_id` dimension", and the paragraph a few lines above now opens with "The cap applies to `client_id` alone". They are saying different things — which labels are capped, versus the ceiling arithmetic — so it is repetition rather than a contradiction, but the two identical openings so close together make the second one read as a restatement and the arithmetic easier to skim past.

Dropping the first clause here would keep the hint doing the one job the paragraph above does not:

> The resulting ceiling is `maxClientIdCardinality + 1`, multiplied by the plan, application and client library combinations an API sees.

callaertanthony (approved):
All three points addressed — the sample line now matches `GatewayNode.APPLICATION_NAME`, the uncapped `client_software_*` dimensions are called out (and the paragraph keeps the reassuring half, that removal-on-last-close still bounds the total by concurrent connections), and the 256 ceiling now reads as suffix-inclusive.

Re-checked at this head: `.version-overrides` untouched, the three tab blocks still balance, lint and the 4.13 GitBook check are green.

One non-blocking nit below — happy for it to be ignored or picked up in a follow-up.

ThibaudAV:
Taken — and it was my paragraph that created the problem, not the hint.

The hint read fine on its own until I opened the paragraph above it with the same clause. Two identical openings that far apart would be fine; this close, the second one reads as a restatement and the arithmetic — the only thing the hint says that the paragraph does not — gets skimmed.

Dropped the first clause, so the hint now does exactly one job:

> The resulting ceiling is `maxClientIdCardinality + 1`, multiplied by the plan, application and client library combinations an API sees.

`lint_overrides.py . apim 4.12 4.13` still passes. Squashed into the single commit.

tolaore (approved):

PR #2287, opened 2026-09-02
docs(apim): document the FIPS images for 4.12 and 4.13

ThibaudAV:
## Issue

https://gravitee.atlassian.net/browse/ESM-109

## Description

The FIPS variants of the APIM images have been published for a while and the documentation never
mentioned them. `chainguard` and `chainguard-fips` appear nowhere in this repository, and image
variants in general are only mentioned in passing — the Alpine/Debian note in
`docker-compose.md`, added for ONNX Runtime, and the default tag note in the Kubernetes guide.

This adds the page that should have been there, for **4.12 and 4.13**, under
`self-hosted-installation-guides/docker/`, linked from the Docker README and registered in each
`SUMMARY.md`.

It covers:

- **What is different inside** — BouncyCastle FIPS as provider 1, BCJSSE as provider 2, `SunJCE`
  and `SunJSSE` absent, approved-only mode.
- **Where to find the images** — the private registry only, never Docker Hub, with the four
  components that publish a FIPS variant.
- **Keystore and truststore formats** — the part that actually catches people out. PEM is the
  only sound option; the page says what happens with the others rather than just recommending.
- **How to configure a PEM keystore**, in Docker and in Helm.
- **Known limitations** — the third-party capabilities that lose their BouncyCastle backing:
  MSSQL certificate authentication and Always Encrypted, the MariaDB `parsec` plugin, encrypted
  and signed PDF handling, and the Argon2/SCrypt encoders.

## Additional context

The two pages are not copies. 4.12 runs on **JDK 21** and 4.13 on **JDK 25**, verified against
the published images (`4.12-chainguard-fips` → 21.0.13, `master-latest-chainguard-fips` → 25.0.4),
and the keystore behaviour differs materially:

| Format | 4.12 (JDK 21) | 4.13 (JDK 25) |
| --- | --- | --- |
| `pem` | works | works |
| `pkcs12` | loads in memory, fails on write | **does not load at all** |
| `jks` | loads, on non-approved SHA-1 PBE | **read-only, does not load** |

So the 4.13 page carries an upgrade warning: a deployment moving from 4.12 while keeping a `jks`
or `pkcs12` keystore will fail to start its TLS listeners.

Every factual claim was measured against the real images rather than inferred — the provider
list, the absence of `SunJCE`/`SunJSSE`, the keystore behaviour per JDK, the registry the images
are pushed to, and which components publish a FIPS variant.

The Helm `certificates` example relies on
https://github.com/gravitee-io/gravitee-api-management/pull/19547, which exposes it for the HTTP
listeners; it already existed for the Kafka listener and the rate limit repository. That PR also
adds the `docker/quick-setup/fips-mtls` stack this page points to, so it should merge first.


docs/apim/4.12/self-hosted-installation-guides/docker/fips-images.md
```diff
@@ -0,0 +1,123 @@
...
+## Keystore and truststore formats
+
+This is the constraint that most often surprises. Under *approved only* mode, the formats APIM
+accepts do not all load:
+
+| Format | On the FIPS images |
+| --- | --- |
+| `pem` | **Works.** The recommended format. |
+| `pkcs12` | Loads in memory, but writing a keystore fails: the PKCS12 key derivation function (`PBEWithHmacSHA256AndAES_256`) has no provider once `SunJCE` is gone. |
+| `jks` | Loads, but relies on SUN's SHA-1 based password encryption, which is not FIPS-approved — using it defeats the purpose of running the FIPS image. |
```
callaertanthony: This contradicts the reproduction stack the page points readers to at the bottom.

`docker/quick-setup/fips-mtls/README.md` on the **`4.12.x`** branch (I diffed it against `master` — byte-identical) says:

> under BC-FIPS *approved only*, PKCS12 key derivation has no provider and JKS is answered read-only, so **PEM is the only format that loads**.

This table says `pkcs12` loads in memory and `jks` loads. Both can't be true for 4.12. Either the page is over-generous about what JDK 21 tolerates, or that README was written for the JDK 25 behaviour and never adjusted when it landed on `4.12.x`.

Worth settling which, since a reader following the "Reproducing an issue" link lands on the contradicting text.

docs/apim/4.12/self-hosted-installation-guides/docker/fips-images.md
```diff
@@ -0,0 +1,123 @@
...
+| `bcfks` | **Works from APIM 4.12.19.** The BouncyCastle FIPS keystore, designed for approved-only mode: unlike PKCS12 it can also be written, so it is the option to reach for when you need a single password-protected container rather than separate certificate and key files. |
```
callaertanthony: The substance checks out, but 4.12.19 isn't released yet.

- `4.12.x` `pom.xml` is at `<revision>4.12.19</revision>` with `gravitee-node 9.9.0`
- gravitee-node's changelog puts BCFKS support in 9.9.0 (`certificates: accept BCFKS as a keystore and truststore type`)
- but the newest tag on the repo is `4.12.18` — there is no `4.12.19`

So if this merges before the release cuts, 4.12 readers are told to reach for a patch they can't pull. Either gate the merge on the 4.12.19 release, or phrase it as upcoming. Same applies to the "On 4.12.19 and later" sentence just below.

docs/apim/4.12/self-hosted-installation-guides/docker/fips-images.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+The tag is the ordinary version tag with the `-chainguard-fips` suffix. It is available for:
+
+| Component | Image |
+| --- | --- |
+| API Gateway | `graviteeio.azurecr.io/apim-gateway:<version>-chainguard-fips` |
+| Management API | `graviteeio.azurecr.io/apim-management-api:<version>-chainguard-fips` |
+| Console | `graviteeio.azurecr.io/apim-management-ui:<version>-chainguard-fips` |
+| Developer Portal | `graviteeio.azurecr.io/apim-portal-ui:<version>-chainguard-fips` |
```
callaertanthony, line 45: CI builds five FIPS images, not four — the Gamma Console is missing from this table.

`.circleci/ci/src/workflows/workflow-build-chainguard-fips-images.ts` has a `Build Gamma Console chainguard-fips docker image` job on **both** `4.12.x` and `master`, using the same `fipsNginxBaseImage` as the Console and Portal.

Either add the row, or say explicitly that the table covers the APIM components only.

docs/apim/4.13/self-hosted-installation-guides/docker/fips-images.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+The tag is the ordinary version tag with the `-chainguard-fips` suffix. It is available for:
+
+| Component | Image |
+| --- | --- |
+| API Gateway | `graviteeio.azurecr.io/apim-gateway:<version>-chainguard-fips` |
+| Management API | `graviteeio.azurecr.io/apim-management-api:<version>-chainguard-fips` |
+| Console | `graviteeio.azurecr.io/apim-management-ui:<version>-chainguard-fips` |
+| Developer Portal | `graviteeio.azurecr.io/apim-portal-ui:<version>-chainguard-fips` |
```
callaertanthony, line 45: Same as on the 4.12 page: the Gamma Console `chainguard-fips` image is built by CI (`workflow-build-chainguard-fips-images.ts` on `master`) but isn't listed here.

docs/apim/4.13/self-hosted-installation-guides/docker/fips-images.md
```diff
@@ -0,0 +1,131 @@
...
+`jks` in some paths. A deployment that upgrades from 4.12 to 4.13 while keeping a `jks` or
+`pkcs12` keystore will fail to start its TLS listeners. Convert those keystores before upgrading —
+to PEM, or to `bcfks` if you would rather keep a single password-protected container, which is the
+closer equivalent of the JKS you are leaving behind.
+{% endhint %}
+
+The JVM-level trust store is a separate matter: `javax.net.ssl.trustStore` does not accept PEM,
+so the images set `-Djavax.net.ssl.trustStoreType=FIPS` to read the bundled `cacerts` in
+compatibility mode. This applies to the TLS connections APIM opens to JDBC, Redis and
+Elasticsearch or OpenSearch.
```
callaertanthony: Where does this come from? I couldn't source it.

Searching `org:gravitee-io` for `trustStoreType` returns zero hits for this flag. It isn't in APIM's `Dockerfile.chainguard`, and it isn't in `gravitee-docker/images/java/Dockerfile.chainguard` either — that file only sets `LANG`/`LANGUAGE`/`LC_ALL` and switches the base to `adoptium-jre-fips`.

So I assume it comes from the upstream Chainguard base image rather than from anything we set. Not saying it's wrong — but "the images set" reads as a Gravitee decision, and if it's inherited from upstream that's worth wording differently, since it can change out from under us.

Same paragraph exists on the 4.12 page at lines 68-70.

docs/apim/4.13/self-hosted-installation-guides/docker/fips-images.md
```diff
@@ -0,0 +1,131 @@
...
+This is the constraint that most often surprises. Under *approved only* mode, the formats APIM
+accepts do not all load:
+
+| Format | On the FIPS images |
+| --- | --- |
+| `pem` | **Works.** The recommended format. |
+| `pkcs12` | **Does not load.** There is no PKCS12 keystore implementation at all on this JDK once `SunJCE` is gone, and requesting one fails immediately. |
+| `jks` | **Does not load.** BouncyCastle FIPS answers for JKS read-only, so APIM cannot open it. |
+| `bcfks` | **Works.** The BouncyCastle FIPS keystore, designed for approved-only mode. Unlike PKCS12 it can also be written, so it is the option to reach for when you need a single password-protected container rather than separate certificate and key files. |
+| `self-signed` | Not recommended: generating the certificate exercises the same non-approved paths. |
```
callaertanthony, line 63: Minor: `pem-folder` is missing from the table. `helm/values.yaml` lists it for truststores — "Supports jks, pem, pkcs12, bcfks, pem-folder (for the latter watch supports added/updated/removed files)".

It matters here because it's in the PEM family the page is already steering people toward, so a reader who wants a watched directory of certs has no entry telling them whether it survives approved-only mode.

docs/apim/4.13/self-hosted-installation-guides/docker/fips-images.md
```diff
@@ -0,0 +1,131 @@
+---
+description: >-
+  What the FIPS variants of the APIM Docker images are, where to pull them, and
+  what they constrain.
+---
+
+# FIPS images
+
+Alongside the ordinary images, APIM publishes a FIPS variant of each backend and frontend
```
callaertanthony: Style nit, both new pages: the prose is hard-wrapped at ~100 characters, while every neighbouring page in this repo leaves paragraph lines unwrapped (`customize-your-installation.md` runs to 555 characters on one line, the docker `README.md` to 1256).

Renders identically either way, so nothing is broken. The cost is that the first edit made through the GitBook UI will reflow the whole file and turn a one-line change into a full-file diff.

callaertanthony (commented):
Reviewed at `279f390`. The writing is genuinely good — the "Known limitations" section and the 4.13 upgrade warning are the kind of thing people hit at 2am. Four things to settle before merge, plus two style nits, all inline.

**The one that matters:** the 4.12 format table and the `fips-mtls` README on the `4.12.x` branch disagree about whether `pkcs12` and `jks` load under JDK 21. The page sends readers to that stack, so they will read both. One of the two needs to change.

### Verified correct

I checked these against the APIM repo at the matching branch, all good:

- JDK 21 (4.12) / JDK 25 (4.13) — `fipsJavaBaseImage` in `.circleci/ci/src/config.ts` is `java:21-chainguard-fips` on `4.12.x` and `java:25-chainguard-fips` on `master`.
- The Docker env vars match `docker/quick-setup/fips-mtls/docker-compose.yml:107-111` character-for-character.
- All five Helm keystore paths exist in `helm/values.yaml`: `gateway.ssl`, `gateway.servers[].ssl`, `gateway.kafka.ssl`, `gateway.ratelimit.redis`, `gateway.distributedSync.redis`.
- `docker/quick-setup/fips-mtls` exists on both `4.12.x` and `master`.
- SUMMARY.md placement and indentation are right in both versions, and the registry really is private-only.

docs/apim/4.12/self-hosted-installation-guides/docker/fips-images.md
callaertanthony: Fixed in `e9ea3a9` — the `Gamma Console` row is there and `gamma-ui` is the right image name (`config.ts:54` on `4.12.x`). Resolving.

docs/apim/4.13/self-hosted-installation-guides/docker/fips-images.md
callaertanthony: Fixed in `e9ea3a9`. Same check on `master`: `gamma-ui`, built from the same `fipsNginxBaseImage` as the Console and Portal. The added sentence about the three UI images serving plain HTTP with TLS terminated upstream also holds — `console-webui/docker/config/default.conf` listens on both ports with no `ssl` directive and no `ssl_certificate`. Resolving.

docs/apim/4.13/self-hosted-installation-guides/docker/fips-images.md
callaertanthony: Answered in `e9ea3a9` — the paragraph now says the FIPS base image carries the flag in `JDK_JAVA_OPTIONS` and that APIM does not set it. That is exactly the attribution I was missing: it reads as inherited from upstream rather than as a Gravitee decision, which is what the search results supported. Resolving.

docs/apim/4.13/self-hosted-installation-guides/docker/fips-images.md
callaertanthony: Fixed in `e9ea3a9`, and the "truststores only" qualifier is correct — I checked. `PEM-FOLDER` is handled solely by `FolderTrustStoreLoaderFactory` in gravitee-node; neither `FileKeyStoreLoaderFactory` nor `FileTrustStoreLoaderFactory` lists it, and there is no keystore equivalent of the folder loader. Resolving. One wording nit on the same row in the follow-up review.

docs/apim/4.12/self-hosted-installation-guides/docker/fips-images.md
callaertanthony: Still open after `e9ea3a9` — and the rewrite made the contradiction sharper rather than smaller, which I think settles which side is wrong.

The failure mode changed between the two revisions:

- before: fails on key derivation, `PBEWithHmacSHA256AndAES_256`
- now: fails on the integrity MAC, `HmacPBESHA256`, quoted verbatim as `calculateMac failed: Algorithm HmacPBESHA256 not available`

`docker/quick-setup/fips-mtls/README.md:9-10` on `4.12.x` (rechecked at the time of writing, unchanged) still says:

> PKCS12 **key derivation** has no provider and JKS is answered read-only, so PEM is the only format that loads.

So the page now says key derivation is fine and the MAC is what breaks, while the README says the opposite. Quoting a real error string makes the page the credible one, so I read the README as the stale text.

Nothing to change in this PR on that account — but it does need a follow-up on `gravitee-api-management` `4.12.x` to fix that README, otherwise anyone following the "Reproducing an issue" link at the bottom of this page lands straight on the contradicting sentence. Leaving this thread open as the pointer to that follow-up; happy for you to resolve it once the other PR exists.

docs/apim/4.12/self-hosted-installation-guides/docker/fips-images.md
callaertanthony: Still open after `e9ea3a9`: the newest tag on `gravitee-api-management` is still `4.12.18`, so `4.12.19` is not out yet. Not asking for a text change — just flagging that this is a merge-timing dependency, since both this row and the "On 4.12.19 and later" sentence below promise a patch nobody can pull today.

callaertanthony (commented):
Re-reviewed at `e9ea3a9`. Four of the six threads are addressed and I have resolved them, each with a note on what I actually checked rather than just taking the change at face value:

- **Gamma Console** — row added, and `gamma-ui` is the correct image name per `config.ts:54` on both branches.
- **`trustStoreType=FIPS`** — now attributed to the base image's `JDK_JAVA_OPTIONS`, with an explicit "APIM does not set it". That was the whole of my question.
- **`pem-folder`** — added, and the "truststores only" qualifier is right: `PEM-FOLDER` is handled solely by `FolderTrustStoreLoaderFactory`, with no keystore equivalent.
- **Hard-wrapping** — gone, matches the neighbouring pages now (you had already resolved that one).

The new sentence about the three UI images serving static content over plain HTTP with TLS terminated upstream also checks out — `console-webui/docker/config/default.conf` listens on both ports with no `ssl` directive and no `ssl_certificate`.

**Two threads left open**, neither of which blocks the text of this PR:

1. The 4.12 `pkcs12`/`jks` rows still disagree with `fips-mtls/README.md` on `4.12.x`. The rewrite actually sharpens the disagreement — see my reply on that thread — and I now think the page is right and the README is stale, so what it needs is a follow-up PR on `gravitee-api-management`, not an edit here.
2. `4.12.19` still has no tag. Merge-timing dependency only.

And two small things introduced by the rewrite, inline below. Both are wording, neither is a correctness problem.

docs/apim/4.12/self-hosted-installation-guides/docker/fips-images.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Keystore and truststore formats
+
+This is the constraint that most often surprises. Under *approved only* mode, the formats APIM accepts do not all load:
+
+| Format | On the FIPS images |
+| --- | --- |
+| `pem` | **Works.** The recommended format. |
+| `pem-folder` | **Works**, truststores only. It goes through the same loader as `pem` rather than a JVM keystore, and `watch` picks up files added, updated and removed in the directory. |
+| `pkcs12` | Loads in memory, but writing a keystore fails: writing one fails on the integrity MAC, which needs `HmacPBESHA256` — an algorithm with no provider once `SunJCE` is gone, surfacing as `calculateMac failed: Algorithm HmacPBESHA256 not available`. |
```
callaertanthony: Duplicated clause from the edit:

> Loads in memory, but writing a keystore fails: **writing one fails** on the integrity MAC, which needs `HmacPBESHA256`

One of the two halves can go. Suggestion:

> Loads in memory, but writing a keystore fails on the integrity MAC, which needs `HmacPBESHA256` — an algorithm with no provider once `SunJCE` is gone, surfacing as `calculateMac failed: Algorithm HmacPBESHA256 not available`.

Separately, thank you for the verbatim error string — that is what convinced me the `4.12.x` `fips-mtls` README is the stale side of the contradiction rather than this table.

docs/apim/4.13/self-hosted-installation-guides/docker/fips-images.md
```diff
@@ -0,0 +1,107 @@
...
+The three UI images are built on a FIPS-validated nginx base rather than a JVM one. They serve static content over plain HTTP, with TLS terminated upstream, so the keystore constraints below concern the Gateway and the Management API only.
+
+## Keystore and truststore formats
+
+This is the constraint that most often surprises. Under *approved only* mode, the formats APIM accepts do not all load:
+
+| Format | On the FIPS images |
+| --- | --- |
+| `pem` | **Works.** The recommended format. |
+| `pem-folder` | **Works**, truststores only. It goes through the same loader as `pem` rather than a JVM keystore, and `watch` picks up files added, updated and removed in the directory. |
```
callaertanthony: Small wording point on "rather than a JVM keystore" — it does end up as a JVM keystore, just not a PKCS12 one.

`KeyStoreUtils` picks the in-memory type by provider:

```java
private static final boolean FIPS_MODE = Security.getProvider("BCFIPS") != null;
public static final String DEFAULT_KEYSTORE_TYPE = FIPS_MODE ? TYPE_BCFKS : TYPE_PKCS12;
```

so the certificates read out of the folder are materialised into a `KeyStore.getInstance(BCFKS)` under FIPS. (Present in gravitee-common 5.2.0, which `4.12.x` pins — so this holds on both versions.)

The distinction the sentence is reaching for is that it is not loaded *from* a keystore file, which is the part that matters for approved-only mode. Something like "it reads the certificates directly rather than opening a keystore file" would say it without the JVM-keystore claim.

Same line on the 4.12 page.

docs/apim/4.12/self-hosted-installation-guides/docker/fips-images.md
ThibaudAV: Fixed in `ae8244d4`, taking your wording. The duplication was mine: I replaced the tail of the sentence when I swapped in the real algorithm name and never re-read the head.

docs/apim/4.13/self-hosted-installation-guides/docker/fips-images.md
ThibaudAV: You are right, and I checked it rather than taking it on trust.

`gravitee-common/src/main/java/io/gravitee/common/util/KeyStoreUtils.java` lines 78-81:

```java
private static final boolean FIPS_MODE = Security.getProvider("BCFIPS") != null;
public static final String DEFAULT_KEYSTORE_TYPE = FIPS_MODE ? TYPE_BCFKS : TYPE_PKCS12;
```

and `KeyStore.getInstance(DEFAULT_KEYSTORE_TYPE)` is called at lines 148, 209 and 244. So the certificates really are materialised into a JVM keystore, a BCFKS one under FIPS. My sentence was wrong, not just loose.

Reworded on both pages in `ae8244d4` to "It reads the certificates directly rather than opening a keystore file, like `pem`" — which keeps the distinction that matters for approved-only mode without the false claim.

docs/apim/4.12/self-hosted-installation-guides/docker/fips-images.md
ThibaudAV: Agreed, and I settled which side is stale by measuring rather than reasoning: I compiled a small probe and ran it inside `graviteeio.azurecr.io/apim-gateway:4.12-chainguard-fips`.

`approved_only=true` confirmed, then per type:

| Type | `getInstance` + `load` | `store` |
| --- | --- | --- |
| `PKCS12` | OK, provider SUN | **fails** — `calculateMac failed: Algorithm HmacPBESHA256 not available` |
| `JKS` | OK, provider SUN | **OK** |
| `BCFKS` | OK, provider BCFIPS | OK |

So on 4.12 / JDK 21 three formats load, PKCS12 is read-only in practice, and JKS is fully writable. The README's "JKS is answered read-only" and "PEM is the only format that loads" are both wrong for this branch — that description matches JDK 25 behaviour, which supports your reading that it landed on `4.12.x` unadjusted.

A follow-up on `gravitee-api-management` is the right place, and it needs different text per branch: on `4.12.x` three formats load and the PKCS12 break is on the integrity MAC, while on `master` `pkcs12` and `jks` do not load at all and `bcfks` now joins PEM since the node 9.9.0 bump landed there.

Happy for this thread to stay open as the pointer until that PR exists.

docs/apim/4.12/self-hosted-installation-guides/docker/fips-images.md
ThibaudAV: Understood, and no text change on our side: the merge will be timed against the 4.12.19 release rather than the wording softened, so the row and the sentence below it become true the moment this lands.

docs/apim/4.12/self-hosted-installation-guides/docker/fips-images.md
callaertanthony: Confirmed fixed in `ae8244d` — the duplicated half is gone and the row now reads as the single sentence. Resolving.

Worth saying that the verbatim error string you put here is what let the other thread get settled by measurement rather than by argument, so it earned its keep twice.

docs/apim/4.13/self-hosted-installation-guides/docker/fips-images.md
callaertanthony: Confirmed fixed in `ae8244d`, on both pages: "It reads the certificates directly rather than opening a keystore file, like `pem`". That keeps the distinction that matters for approved-only mode without the JVM-keystore claim. Resolving.

tolaore (approved):

PR #2309, opened 2026-09-03
docs(gamma): document environment-scoped shared policy groups (FOUND-4)

tolaore:
https://gravitee.atlassian.net/browse/FOUND-4


ankita-gupta21:
Looks good to me

PR #2331, opened 2026-09-04
docs(gamma): LLM Proxy Failover page (AIAM-96)

tolaore:
https://gravitee.atlassian.net/browse/AIAM-96


araffass (approved):
Conflicts solved - LGTM

PR #2333, opened 2026-09-04
docs(apim): document the HTTP/2 flow-control window settings

thomasballivet:
**Issue**

https://gravitee.atlassian.net/browse/APIM-15040

**Description**

Adds a `Tune HTTP/2 flow-control windows` subsection to the existing `Enable HTTP/2 support`
section of the Gateway HTTP server page, on **4.9, 4.10, 4.11, 4.12 and 4.13** — the versions
that ship the two options, matching the backport plan of
[gravitee-node#607](https://github.com/gravitee-io/gravitee-node/pull/607).

Upload throughput over HTTP/2 is bounded by `window size / round-trip time`, so the 65535
bytes protocol default throttles large request bodies as soon as there is real network
latency, while HTTP/1.1 is unaffected. A customer hitting that symptom looks at this page,
which today stops at `alpn: true`.

**The rule the page has to carry**

Both options have to be set. Vert.x resizes the connection window only on a strictly positive
value, and the initial settings the Gateway sends reach streams only, never the connection, so
raising `streamWindowSize` alone changes nothing observable. That is stated as a warning hint,
along with the memory cost of raising them.

Each `Helm values.yaml` tab shows both deployment shapes: `gateway.http.http2` for the single
server the chart renders when `gateway.servers` is unset, and `gateway.servers[].http2`
otherwise, since the chart takes the HTTP server options from that list and ignores the rest of
`gateway.http`. Leading with the form the surrounding page already uses.

**Dependency — lifted**

The APIM side is merged on the five lines, each carrying the chart change
(`helm/templates/gateway/gateway-configmap.yaml`, both shapes) and the `gravitee-node` bump
that makes the Gateway read the options: gravitee-io/gravitee-api-management#19653 (master),
gravitee-io/gravitee-api-management#19685 (4.12.x), gravitee-io/gravitee-api-management#19686 (4.11.x),
gravitee-io/gravitee-api-management#19687 (4.10.x), gravitee-io/gravitee-api-management#19688 (4.9.x).

The options therefore arrive in **APIM 4.13.0, 4.12.20, 4.11.28, 4.10.31 and 4.9.35**, none of
which is released yet. Worth deciding at merge time whether these pages publish now or once
those versions ship — a reader on an earlier patch sets a value their Gateway ignores.


phiz71 (commented):
Reviewed alongside gravitee-io/gravitee-node#607 and gravitee-io/gravitee-api-management#19653.

The content is technically right — I checked the central claim against Vert.x 5.0.12 and Netty 4.2.17, and it holds: `HttpServerConnectionInitializer` resizes the connection window only on a strictly positive `http2ConnectionWindowSize`, and the initial settings reach Netty's `DefaultHttp2LocalFlowController.initialWindowSize`, which walks `forEachActiveStream` and never touches stream 0. So "raising `streamWindowSize` alone changes nothing observable" is exact, not an approximation, and it deserves the warning hint it got.

Also worth saying: the heading style is matched per page — bold on 4.9 and 4.10, plain on 4.11 through 4.13, following each page's own convention. That is the detail that normally gets copy-pasted wrong across five files.

Two things to sort out before this goes live, both about the surrounding context rather than the prose.

## 1. The Helm tab documents a form that half the deployments cannot use

`gateway.http.http2` is rendered by the chart **only when `gateway.servers` is unset**. In `helm/templates/gateway/gateway-configmap.yaml`, the `http:` block that carries `websocket`, `haproxy` and now `http2` sits in the `{{- else }}` branch of `{{- if .Values.gateway.servers }}`; the multi-server branch takes its values from `$server.http2` instead. The chart supports both — gravitee-io/gravitee-api-management#19653 adds both, and its CHANGELOG entry says so explicitly ("also per server under `gateway.servers[]`") — but this page only shows one.

That is sharpest on the **4.9 page**, whose existing `Enable HTTP/2 support` section shows `alpn` in exactly the other form:

```yaml
gateway:
  servers:
    - type: http
      alpn: true
```

A reader who follows that example, then scrolls down to the new subsection and copies `gateway.http.http2`, sets a value the chart silently drops. Which is, almost exactly, the failure mode the subsection above it warns about.

Worth showing the `gateway.servers[]` variant too, at least on the pages whose Helm examples already use it.

## 2. The pages go live before any release accepts the options

Today: APIM `master` and `4.12.x` are on `gravitee-node` 9.8.0, `4.11.x` on 8.3.0, `4.9.x` and `4.10.x` on 7.27.0 — and gravitee-node#607, which introduces the two options, is not merged yet. So as soon as this lands, five published pages describe `gravitee.yaml` and `.env` settings that no released Gateway reads, on top of a Helm key whose chart change exists on `4.12.x` alone.

The PR body flags the chart dependency, which is the narrower one. The node dependency applies to all three tabs and to every version here. Either hold this until the node bump reaches each line, or name the minimum patch version on each page — the second is more work but survives someone landing the branches out of order.

## Smaller things

- The 4.9 page titles its three existing Helm tabs `Helm values.yml`; the new one is `Helm values.yaml`. Same page, two spellings.
- The 4.10 page's `Enable HTTP/2 support` section has no tabs at all — a bare YAML block — and the page has no `.env` example anywhere. The new subsection introduces a full three-tab block. It reads as an improvement rather than a mismatch, just noting it is a change of shape and not a match to the surrounding page.

---

# Mergeable?

**Yes, once sequenced.** Nothing in the prose needs rewriting; what needs settling is when this publishes and whether the Helm tab covers both deployment shapes.

### MUST
Nothing in the content.

### SHOULD

1. **Document the `gateway.servers[]` form as well** — the chart supports it, and on the 4.9 page it is the form the section directly above already uses.
2. **Sequence the merge behind the `gravitee-node` bump** in each APIM line, or state the minimum version on each page. Right now no released Gateway accepts these options.

### COULD

3. Align the new Helm tab title with the page's existing `Helm values.yml` on 4.9.
4. On 4.10, consider whether adding a three-tab block to a section that has none is the shape you want, or whether the section as a whole should move to tabs.

thomasballivet:
Both SHOULD, 2 COULD out of 2. Pushed as a73108e; PR set to draft.

**Taken**

- **1** — Checked the template branch before writing it, and it is exactly as you say: the `http:` block carrying `alpn`, `websocket`, `haproxy` and now `http2` is the `{{- else }}` of `{{- if .Values.gateway.servers }}`, and with a server list `http:` keeps only `requestTimeout`, `requestTimeoutGraceDelay` and `pathHandling`. A rendered `values.yaml` with `gateway.servers` set confirms `gateway.http.http2` is dropped. The Helm tab now shows both forms on the five pages, with the sentence saying which one applies, ordered per page: `servers[]` first on 4.9 where the `alpn` example right above uses it, single-server first elsewhere for the same reason.
- **3** — `Helm values.yml` on 4.9, matching the two other Helm tabs of the page.
- **2** — Draft until the sequencing is done. Naming a minimum version is not possible yet: gravitee-node#607 is unmerged, so the patch that will carry the options has no number. And the dependency is one notch wider than the review says — the Helm tab, new `servers[]` variant included, is only true on **4.12.x**, since gravitee-io/gravitee-api-management#19653 targets that branch alone. So: node#607 merged, node bumped in each line (master/4.12.x on 9.8.0, 4.11.x on 8.3.0, 4.9/4.10.x on 7.27.0), chart backported to 4.9.x/4.10.x/4.11.x, then this publishes.

**Kept as is**

- **4** — The three tabs stay on 4.10. The page already uses `{% tabs %}` for the three keystore variants, so the shape is not foreign to it; the `.env` tab is indeed the only one on that page. Happy to cut it back to a bare `gravitee.yaml` block if you would rather match the section above it, but that drops Helm and `.env` where they are the ones that matter.

garethbrinn (approved):

PR #2350, opened 2026-09-08
docs(gamma): manage platform policies from the organization's Policy Studio (FOUND-73)

tolaore:
https://gravitee.atlassian.net/browse/FOUND-73


docs/gamma/4.13/platform-management/gamma-release-notes.md
```diff
@@ -123,7 +123,7 @@ Event Stream Management adds a duplication path for Kafka Services.
...
+Platform Management adds environment-scoped dictionaries and metadata as reusable assets for APIs and API policies, gateway routing configuration for the organization, and organization-wide user administration. Tenants pair each gateway with the endpoints it loads. Groups collect the users of an environment behind shared default roles, and shared policy groups bundle policy steps for reuse across API flows. Platform flows apply policies to every API in the organization. It also adds a view of the gateway instances running behind an environment, and an audit trail of configuration changes at both organization and environment scope. It adds the organization-wide console settings too, covering console authentication, console behavior, cross-origin access to the Management API, and outbound email. Custom observability dashboards gain server-side storage.
```
jourdiw:
```suggestion
Platform Management adds environment-scoped dictionaries and metadata as reusable assets for APIs and API policies, gateway routing configuration for the organization, and organization-wide user administration. Tenants pair each gateway with the endpoints it loads. Groups collect the users of an environment behind shared default roles, and shared policy groups bundle policy steps for reuse across API flows. Platform flows apply policies on request and response phases to every API in the organization, except Native Kafka APIs as they do not have those phases. It also adds a view of the gateway instances running behind an environment, and an audit trail of configuration changes at both organization and environment scope. It adds the organization-wide console settings too, covering console authentication, console behavior, cross-origin access to the Management API, and outbound email. Custom observability dashboards gain server-side storage.
```

tolaore: Applied by hand in 09ac4235a, thanks. Note the suggestion was written before the rebase, so applying it with the button would also have dropped the environment-alerts sentence that arrived in this paragraph from APIM-14816 — I kept both.

Both halves of your point check out against the gateway:

* Phases: the organization chain is only ever invoked on `REQUEST` and `RESPONSE` (`DefaultApiReactor:360,388`, `SyncApiReactor:266,285`). `MessageApiReactor` also calls it on the message phases, but `HttpPolicyChainFactory.getSteps` maps only `REQUEST` to `pre` and `RESPONSE` to `post`, so those chains resolve empty.
* Native Kafka: `KafkaApiReactor` has no organization chain, and `NativeFlow` carries only `entrypointConnect`, `interact`, `subscribe`, and `publish`.

Two things I added on top of your wording, both worth a sanity check from you:

1. **TCP proxy APIs are a second exception.** `TcpApiReactor.handle` runs entrypoint request, invoker, entrypoint response, with no flow chain at all, and the console disables the Policies entry when an API has a TCP listener (`api-v4-menu.service.ts:363-374`). The reason differs from native Kafka's, so each gets its own clause rather than sharing "don't have those phases".
2. **v4 message APIs run platform flows**, so I added them to the "Understand where platform flows run" list on the page (`MessageApiReactor:148,184`).

The paragraph now reads:

> Platform flows apply policies on request and response phases to every API in the organization. Native Kafka APIs don't have those phases, and TCP proxy APIs don't run policy flows, so both are left untouched.

Split into two sentences to stay under the Vale sentence-length rule, and "do not" became "don't" for `Microsoft.Contractions`. Happy to drop the TCP clause if you'd rather keep the scope to Kafka.

garethbrinn (approved):

garethbrinn (approved):

PR #2376, opened 2026-09-15
New Gamma overview page

garethbrinn:


samprodger (approved):
Approving. The outcome-to-feature tables read far better than the old prose lists, and the new Observability section fills a real gap. Relative cross-references throughout are a solid improvement. I checked the figure paths resolve on both 4.12 and 4.13, they do.

One thing I would like back, @garethbrinn, and it is not blocking.

The MCP Proxy's two modes have gone from the overview. The old page had "Proxy mode for transparent governance of upstream MCP servers, and Studio mode for composing catalog assets into an MCP Studio".

That distinction is the strongest differentiator we have on the MCP side. Plenty of gateways can proxy an MCP server or bring one in. Very few can compose tools from several upstreams into a single governed endpoint and then layer tool-level access policy over the top so different personas see different toolsets. When I demo it, that is the moment the room changes. An overview that does not mention it leaves the reader thinking we do what everyone else does.

A line in the Agent Management row of the modules table, or in the "Expose existing infrastructure to AI agents" section next to the Create an MCP proxy link, would cover it.

Two smaller notes:

- "microsecond latency with no network hop" appears twice as a flat claim. Worth either sourcing it or softening it, since it is the kind of number a prospect repeats back at us.
- Vale is reporting `skipping` rather than passing, so style linting has not actually run on this. Probably worth a glance at why before it becomes the norm.

samprodger:
@garethbrinn approved

samprodger:
needs rebasing

tolaore (approved):

